PR #109, opened 2011-09-01
Allow ezuser datatype fromString to be used for updates

peterkeung:
When creating a user, it's important to make sure there isn't an existing user with the same username or e-mail address (when RequireUniqueEmail is enforced).  However, when updating a user, this limitation should be lifted if the match is on the user that you're trying to update.


skrosoft:
It seems there is a problem with the serialized draft present in the data_text attribute. The fromString function is overrided with the content/publish

PR #371, opened 2012-07-16
Fixed #19174: previous versions XMLText embeds may not be added as relation

bdunogier:
This fix is similar to one that was introduced earlier, that scans other languages of published ezxmltext attributes, and makes sure external links are registered in the database.

Without it, embedded relations that only exist in languages other than the published one won't be registered in ezcontentobject_link. I have thoroughly checked the eZXML API, and there was no cleaner way to get the list of embeds.

The pull request isn't complete yet, it misses:
- unit tests
- a fix script

Those should come today, and I'll push 'em to the same branch

### DIFF
--- a/update/common/scripts/5.0/restorexmlrelations.php
+++ b/update/common/scripts/5.0/restorexmlrelations.php
@@ -23,8 +23,7 @@ $script = eZScript::instance(
 
 $script->startup();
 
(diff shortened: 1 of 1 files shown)


pkamps:
Just a note: the publishing operation is updating the object relations but it's not working in a multi- lingual setup. The code changes in the pull request are fixing the issue. Still, you could consider to fix the publishing operation instead.

PR #435, opened 2012-09-13
Not only tags should be stripped from the login, but quotes are danger t...

SerheyDolgushev:
...oo. Users with danger login should be renamed, but not disabled


SerheyDolgushev:
@glye seems like closing this PR is the wisest move :)

PR #523, opened 2012-12-12
Allows for $UserParameters in content/search

harmstyler:
Custom `$view_parameters` are not currently allowed Search, this update allows for custom view parameters by checking for `$Params['UserParameters']` and merging the array with the existing `$view_parameters` array.


harmstyler:
This probably isn't need anymore. Closing.

PR #551, opened 2013-01-28
Fixed EZP-20301: FinishUserRegistration event has no check for object ve...

xc:
Fix for https://jira.ez.no/browse/EZP-20301#comment-70545.

It only affects admin. The frontend is fine.


petrmifek:
I can confirm this is still an issue with latest master and the proposed change is good enough to resolve it. Thanks.

PR #637, opened 2013-05-15
eZDFS: don't throw exception with empty string as path

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-23409

If a path does not exist in in eZDFSFileHandlerDFSBackend, do not pass an empty string (realpath) to thrown exception.

### DIFF
--- a/kernel/private/classes/clusterfilehandlers/dfsbackends/dfs.php
+++ b/kernel/private/classes/clusterfilehandlers/dfsbackends/dfs.php
@@ -14,16 +14,16 @@ class eZDFSFileHandlerDFSBackend implements eZDFSFileHandlerDFSBackendInterface
     {
         $mountPointPath = eZINI::instance( 'file.ini' )->variable( 'eZDFSClusteringSettings', 'MountPointPath' );
 
-        if ( !$mountPointPath = realpath( $mountPointPath ) )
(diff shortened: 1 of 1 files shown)


bdunogier:
Do we want to merge this ? Given that master will never be released officially...

brookinsconsulting:
Please don't give up on the legacy users who might need improvements now and in the future. It is painful to hear comments like, "Do we want to merge this ? Given that master will never be released officially...".

bdunogier:
Well, we know that we (eZ) will not release another major version of ezpublish legacy. It doesn't affect maintenance ones, of course.

andrerom:
@bdunogier my +1 still stands, do you have any concerns for regressions?

bdunogier:
Hmmm... I guess not. I'm not -1 :-)

andrerom:
@bdunogier lets maybe turn the question around, is this something we should backport? hence which should defiantly go into master first :)

bdunogier:
I don't think it adds any risk, and it could improve experience. I'm not against a backport.

If we don't backport it, there is no point in merging it, is there ? :-)

andrerom:
kind of, unless we make it part of 2015.01 or future tags of course.

PR #735, opened 2013-08-28
Add "cluster/storeFile" cluster event

jeromegamez:
I saw that `cluster/deleteFile` is called in `eZDFSFileHandlerMySQLiBackend::_store()` and thought this couldn't be the intended use.

Then I thought it could be a good idea to introduce a new cluster event that notifies listeners on file creation, as a complement to `cluster/deleteFile`.

So here it is :).

Cheers
:octocat: Jérôme

### DIFF
--- a/doc/features/4.7/cluster_events.md
+++ b/doc/features/4.7/cluster_events.md
@@ -98,6 +98,9 @@ Try to use constants instead or configuration objects passed to your custom gate
 - `cluster/storeMetadata` - **Notification**.
   The metadata array is passed as argument.
 
+- `cluster/storeFile` - **Notification**.
+  The file path, datatype and scope are passed as arguments.
+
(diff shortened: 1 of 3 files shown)


jeromegamez:
@andrerom @bdunogier Gentle ping

bdunogier:
Argh...

I'm now thinking that we might have used delete here for an actual reason... we were iirc implementing some cache thing using those events, and we needed to clear cache when something was stored, to make sure a fresh version would be grabbed upon the next request.

It's ugly, but I'm now thinking this might have consequences...

jeromegamez:
I couldn't find a place where this event has been used, but could a solution then be to register an event listener that listens to the store event and does the same as the delete event?

The thing is, what if there is another event listener for the delete event that destroys something, although in this case, we are clearly creating something, even if it's just e.g. a logger that logs cluster events?

PR #743, opened 2013-08-30
Fix EZP-21477: URL wildcards with leading slash are not matched

jeromegamez:
At /content/urlwildcards, when the user creates an URL wildcard alias starting with a slash, it will not get matched.

I think the easiest solution would be to trim leading slashes from the wildcard when the user enters it, so here it is.

While at it, I also added a `/** @var eZModule $Module */` at the top of the file to satisfy IDEs and fixed a possible E_NOTICE for a possibly undefined `$limitValues` variable and a missing index. PHPStorm code inspection rules :).

https://jira.ez.no/browse/EZP-21477

Cheers
:octocat: Jérôme

### DIFF
--- a/kernel/content/urlalias_wildcard.php
+++ b/kernel/content/urlalias_wildcard.php
@@ -6,6 +6,7 @@
  * @package kernel
  */
 
+/** @var eZModule $Module */
 $Module =& $Params['Module'];
 $http = eZHTTPTool::instance();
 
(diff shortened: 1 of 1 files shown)


jeromegamez:
Gentle ping @bdunogier @patrickallaert @andrerom (Blame github's new PR overview page :) )

patrickallaert:
I change my previous +0.5 to a +1 :) Time for merge!

jeromegamez:
Hurra! <3

andrerom:
+1

PR #769, opened 2013-09-24
Fix for EZP-21605: Version history Status column is missing data

peterkeung:
Since we have 2 newer object version statuses, the Manage Versions interface should show information about these new statuses.

Link: https://jira.ez.no/browse/EZP-21605


bdunogier:
+1 with cleanup as well

peterkeung:
Replaced with a clean pull request; same diff: https://github.com/ezsystems/ezpublish-legacy/pull/1132

PR #789, opened 2013-10-14
Constructor refactoring (PHP4 to PHP5 __construct())

jeromegamez:
Due to eZ Publish's history, most of the classes use PHP4 style constructors in the form

``` php
class MyClass
{
    function MyClass()
    {
        // Constructor
        $this->MyParentClass();
        // More code
    }
}
```

instead of

``` php
class MyClass
{
    public function __construct()
    {
        // Constructor
        parent::__construct();
        // More code
    }
}
```

In our modern times, this is problematic due to the following reasons:
- If you want to use the parent's constructor, you have to call it explicitely
- Derived classes call the parent constructor by its name instead of `parent::__construct()` which makes the refactoring of class names cumbersome
- We don't know if and when a future version of PHP will stop supporting the PHP4 constructor methods

This pull request changes this by doing the following:
- Replace every PHP4 style constructor with the PHP5 style `public function __construct()`
- Replace every call to the parent's constructor with `parent::__construct()`
- Delete all empty constructors where it makes sense (e.g. when the parent has no constructor or overriding the parent's constructor wouldn't make sense)
- Delete all constructors that just call the parent constructor
- For the more "popular" classes that are most likely to be used in 3rd party extensions, the new constructor is just added, and the old style constructor is marked as deprecated (although I personally would prefer not to do that, 3rd party extensions shouldn't be too old and updated)
- Replace every other direct call of the old style constructor

If you think more classes should keep the old style constructor (marked as deprecated) in addition to the new one, please let me know.

Cheers
:octocat: Jérôme

PS: The Travis build failed, but only on one node. I expect the fail not to be related to the PR.


jeromegamez:
@patrickallaert @lolautruche @gggeek @pkamps 

In light of PHP7 being released and @patrickallaert's https://github.com/patrickallaert/ezpublish-legacy-php7, would you perhaps be open to reconsider this PR? I just rebased it with the current master.

Of course, BC is still an issue, but by now, this wouldn't be a [premature optimization](#issuecomment-26416819) any more, but required to be able to run the old stack on PHP 7.

jeromegamez:
I also cherry picked @patrickallaert's [exception handler change](https://github.com/patrickallaert/ezpublish-legacy-php7/commit/ba26b67a9afce5395cec7266fb97ee86b0e60300).

jeromegamez:
This PR can be closed again, as it is superseded by #1233

PR #819, opened 2013-10-27
store can not be called here, because it clears object's data_map

SerheyDolgushev:
For example we have a content class, which consist of following attributes:
- attribute 1 (any datatype)
  ...
- attribute 10 (any datatype)
- attribute 11 (objectrelation list datatype)
- attribute 12 (any datatype)
  ...
- attribute 20 (any datatype)

When new content object of this type is published (without any relations in attribute 11). It`s datatype cache will be cleared. And it will be not possible to get valid object`s datamap in attribute 12 ... attribute 20.

Current changes fixes this bug.

### DIFF
--- a/kernel/classes/datatypes/ezobjectrelationlist/ezobjectrelationlisttype.php
+++ b/kernel/classes/datatypes/ezobjectrelationlist/ezobjectrelationlisttype.php
@@ -200,7 +200,6 @@ class eZObjectRelationListType extends eZDataType
         {
             $content['relation_list'] = array();
             $contentObjectAttribute->setContent( $content );
-            $contentObjectAttribute->store();
             return true;
         }
 


SerheyDolgushev:
@glye 
> Did you verify that only this datatype does this?
To be honest, I do not remember, but most likely "yes" is the answer

> Is the docblock wrong too, then?
Seems like it is

At any rate, is there a chance this PR will be merged? Or should we just wisely close it?

glye:
It seems sensible. But finding the time to verify it is the problem. We need to at least compare with other datatypes.

glye:
Now I feel bad 😃 I had a look at some other datatypes, and the PR seems correct. If you reopen it I will merge.

SerheyDolgushev:
I was unable to reopen this one, as the source branch was removed some time ago. So I made https://github.com/ezsystems/ezpublish-legacy/pull/1455.

PR #827, opened 2013-11-07
Fix EZP-21913: ezpublish:legacy:script break with usage of globals in scripts

lolautruche:
https://jira.ez.no/browse/EZP-21913
https://jira.ez.no/browse/EZP-21941

Replaces #826 .
Complementary PR: #831

This PR replaces usage of flawed `global` keyword by closures. This allows to _import_ context from inside the script.
Reason: These variables are not global anymore when scripts are executed via ezpublish:legacy:script.
## Scripts to fix up
- [x] changelog2xmltext.php
- [x] clusterize.php
- [x] convertprice2multiprice.php
- [x] ezconvertdbcharset.php
- [x] ezflowupgrade.php
- [x] ezwebincommon.php
- [x] ezwebininstall.php
- [x] ezwebinupgrade.php

### DIFF
--- a/bin/php/convertprice2multiprice.php
+++ b/bin/php/convertprice2multiprice.php
@@ -23,9 +23,6 @@
 // script initializing
 require_once 'autoload.php';
 
-global $cli;
-global $currencyList;
-
 $currencyList = false;
 
 $cli = eZCLI::instance();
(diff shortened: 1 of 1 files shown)


lolautruche:
@andrerom @bdunogier Should this be merged or closed ?

andrerom:
I assume this is still an issue, so with mode change and eZDir fix we should merge this.

PR #1005, opened 2014-06-24
EZP-23031: allow large file sizes on DFS without memory limitation.

joaoinacio:
This updates `eZFile::createFile()` so it can accept either a string or file resource as `$data` parameter.
dfs.php is updated accordingly, creating a handle using fopen() instead of passing the whole content with `file_get_contents()`.

The validity check for `$contents` in `createFile()` also applies to the handle in the same way.


bdunogier:
Fine by me.

andrerom:
+1

PR #1012, opened 2014-07-01
Fix EZP-23088: Image information still displayed after removal

yannickroger:
Link: https://jira.ez.no/browse/EZP-23088
## Description

This follows #1001 were deletion of images has been fixed (see description for context for this PR).
`contentObjectAttributes` are modified during the fetch (it is were deletion is done):
https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/content/attribute_edit.php#L288

but the variable is not updated.

This patch refreshes the variable, this way updated content is displayed on the edit page.
## Tests

Manual tets


giRobert:
This change seems to cause a bigger side-effect: https://jira.ez.no/browse/EZP-29723

PR #1099, opened 2014-10-28
Fix: EZP-23531 - ezpublish-legacy composer.json pointing to ~5.3 instead of dev-master

pedroresende:
#### EZP-23531 - ezpublish-legacy composer.json pointing to ~5.3 instead of dev-master

The https://github.com/ezsystems/ezpublish-legacy/blob/master/composer.json file points ~5.3 on the legacy ezsystems extensions when it should be poiting to dev-master

```
        "ezsystems/ezautosave-ls": "~5.3",
        "ezsystems/ezdemo-ls-extension": "~5.3",
        "ezsystems/ezflow-ls-extension": "~5.3",
        "ezsystems/ezgmaplocation-ls-extension": "~5.3",
        "ezsystems/ezie-ls": "~5.3",
        "ezsystems/ezmbpaex-ls": "~5.3",
        "ezsystems/ezmultiupload-ls": "~5.3",
        "ezsystems/ezodf-ls": "~5.3",
        "ezsystems/ezprestapiprovider-ls": "~5.3",
        "ezsystems/ezscriptmonitor-ls": "~5.3",
        "ezsystems/ezsi-ls": "~5.3",
        "ezsystems/ezstarrating-ls-extension": "~5.3",
        "ezsystems/ezwt-ls-extension": "~5.3",
        "ezsystems/ezfind-ls": "~5.3|>=2014.05",
```


andrerom:
Ok, but then I think we can close this as this would be done by @pedroresende in root project.

@lolautruche the challenge is that when using minimum-stability you also get unstable packages from third party packages (bundles/libs), so QA might get issues we don't see as we use "stable" versions of those.

gggeek:
Sure to be OT as ever, but remember to take outmost care to adjust dependency versions when tagging (in the sense of locking them down to what is current at time of release).
I have recently been spending time with a dev who was trying to install a not-current CP version (one or two revisions behind the latest), and the dependencies were not working because they were too open to forward-in-time versions.
I can try to retrieve the exact details and open an issue in Jira if deemed useful.

PR #1123, opened 2014-12-04
Ezp 23753 node alias translation

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-23753

Problem:
When a node that exists on a language different from the siteaccess default language, but has a translated parent/path, the UrlAlias will be incorrect.
For example, for a "FreFolder/EnArticle" the returned alias in fre siteaccess is "EnFolder/EnArticle".
Besides incorrect, this causes additional problems when there is a `PathPrefix` setting for that SA.

This happens because `eZContentObjectTreeNode` requests the alias in the current node's language:
``` eZURLAliasML::fetchPathByActionList( "eznode", $pathArray, $this->CurrentLanguage );```

The fix from EZP-20494 (PR #572) resolved this by prefering the wanted language ONLY for the last node, and returning the rest of the path according to the prioritized languages, however it introduced a change in behavior as well.

This PR resolves the problem by reverting 20494 (if applicable), and having the node fetch the alias full path using solely the prioritized language list.

### DIFF
--- a/kernel/classes/ezcontentobjecttreenode.php
+++ b/kernel/classes/ezcontentobjecttreenode.php
@@ -1883,7 +1883,7 @@ class eZContentObjectTreeNode extends eZPersistentObject
         {
             return $limitation;
         }
-        
+
         $currentUser = eZUser::currentUser();
         $currentUserID = $currentUser->attribute( 'contentobject_id' );
         $limitationList = array();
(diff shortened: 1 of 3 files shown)


yannickroger:
ping @joaoinacio, you can merge it

PR #1131, opened 2014-12-23
Fix EZP-23836 Translations of templates strings in eng-GB needed

ernestob:
Forcing translations when the locale is eng-GB

Adding a new setting to the regional settings in the site.ini
to allow the user to force eZ Publish to use translations
when the locale is set to eng-GB

### DIFF
--- a/settings/site.ini
+++ b/settings/site.ini
@@ -905,8 +905,10 @@ ContentObjectLocale=eng-GB
 # If enabled, ShowUntranslatedObjects will force showing the objects
 # existing in other languages than those specified in SiteLanguageList
 ShowUntranslatedObjects=disabled
-# This setting allow us to translate eng-GB when the default locale is different
-# that eng-GB
+# eZ Publish assumes that eng-GB is the base locale. When the default locale is
+# something other than eng-GB, setting this to "enabled" ensures that eZ Publish
+# will look for an eng-GB translation file.
+# The eng-GB translation is usually always available.
 ForceEngGBTranslation=disabled
 # Prioritized list of languages. Only objects existing in these
 # languages will be shown (unless ShowUntranslatedObjects is enabled).


dougplant:
I like the fix, but I don't understand the comment in the INI ... "This setting allow us to translate eng-GB when the default locale is different that eng-GB"

peterkeung:
This just came up on another site.

Ping @dougplant @thiagocamposviana @pkamps @bgamrat @benkmugo for review.

peterkeung:
Maybe it's more accurate to describe it as:
eZ Publish assumes that eng-GB is the base locale. When the default locale is something other than eng-GB, setting this to "enabled" ensures that eZ Publish will look for an eng-GB translation file.

dougplant:
Yep. "The eng-GB translation is usually always available." might complete it. 

Either way, it is a useful fix.

benkmugo:
I like the fix as it avoids using a work-around or other kind of 'hack'.
Also there's very little to it in terms of loc / logic.

I think the combination of Peter's and Doug's setting description as:
> eZ Publish assumes that eng-GB is the base locale. When the default locale is something other than eng-GB, setting this to "enabled" ensures that eZ Publish will look for an eng-GB translation file.
> The eng-GB translation is usually always available." 

should cover what this does quite clearly.

brookinsconsulting:
We also would very much like this improvement to be merged.

The ini description comment improvement is also a wise idea.

Cheers,
Brookins Consulting

kernel/common/ezpi18n.php
```diff
@@ -109,13 +109,14 @@ public static function tr( $context, $source, $comment = null, $arguments = null
     protected static function translateText( $context, $source, $comment = null, $arguments = null )
     {
         $localeCode = eZLocale::instance()->localeFullCode();
-        if ( $localeCode == 'eng-GB' )
+        $ini = eZINI::instance();
+        $forceEngGBTranslation = $ini->variable( 'RegionalSettings', 'ForceEngGBTranslation' ) === 'enabled';
+        if ( $localeCode == 'eng-GB' && !$forceEngGBTranslation )
```
bdunogier: `===` please :-)

bdunogier:
It looks okay to me.

Maybe it'd be clearer/cleaner if the 3 lines were moved to a method, and used there. It would be more readable, as the method name could replace the comment:

`if ( self::translateStrings( $localeCode ) ) {`

Not a _must_ though.

PR #1132, opened 2015-01-07
Fix for EZP-21605: Version history Status column is missing data

peterkeung:
Since we have 2 newer object version statuses, the Manage Versions interface should show information about these new statuses.

Link: https://jira.ez.no/browse/EZP-21605

Cleaned up version of https://github.com/ezsystems/ezpublish-legacy/pull/769


yannickroger:
+1

andrerom:
+1 (including backports)

yannickroger:
Thanks @peterkeung

PR #1133, opened 2015-01-08
Fix EZP-23870: async publishing caused stale expiry timestamps.

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-23870

With async publishing script, expiry.php data is being loaded at the start and saved at every publish (through script-shutdown of the forked child process.

Because async publishing is a long-running script/daemon, some expiry keys are not modified and end up overwriting possibly newer timestamps (for example from a cache-clear command).

This resolves it by re-loading expiry timestamps before the publish operation.

### DIFF
--- a/kernel/private/classes/ezpcontentpublishingqueueprocessor.php
+++ b/kernel/private/classes/ezpcontentpublishingqueueprocessor.php
@@ -33,6 +33,9 @@ class ezpContentPublishingQueueProcessor
         $this->cleanupInterval = $this->contentINI->variable( 'PublishingSettings', 'AsynchronousCleanupInterval' );
         $this->cleanupAgeLimit = $this->contentINI->variable( 'PublishingSettings', 'AsynchronousCleanupAgeLimit' );
 
+        // initiate current time for cache refresh
+        $this->lastRefreshTimestamp = time();
+
         // Queue reader handler
         $this->queueReader = $this->contentINI->variable( 'PublishingSettings', 'AsynchronousPublishingQueueReader' );
         $reflection = new ReflectionClass( $this->queueReader );
@@ -109,6 +112,8 @@ class ezpContentPublishingQueueProcessor
                     'async.log'
                 );
 
+                $this->refreshCacheStateInfo();
+
                 $pid = $publishingItem->publish();
                 pcntl_signal( SIGINT, $this->signalHandler );
                 $this->currentJobs[$pid] = $publishingItem;
@@ -136,6 +141,22 @@ class ezpContentPublishingQueueProcessor
                 sleep( 1 );
             }
         }
+    }
+
+    /**
+     * Refreshes/reloads cache and other state information periodically.
+     */
+    private function refreshCacheStateInfo()
+    {
+        if ( time() < ( $this->lastRefreshTimestamp + $this->sleepInterval ) )
+        {
+            return;
+        }
+
+        // Reload expiry timestamps
+        eZExpiryHandler::instance()->restore();
+
+        $this->lastRefreshTimestamp = time();
     }
 
     /**
@@ -352,6 +373,12 @@ class ezpContentPublishingQueueProcessor
      * @var int
      */
     private $cleanupLastTime;
+
+     /**
+     * Time counter for refresh of cache/state information
+     * @var int
+     */
+    private $lastRefreshTimestamp;
 
     /**
      * Interval for cleanup of finished processes. Default value is 12 hours in seconds


joaoinacio:
Actually, i am not certain if the restore should be done on each fork()'ed child, or in the main loop of the queue processor with additional "timing" logic.
Reasoning: if there are `n` processes running "in parallel" (depending on slots), it could save a few unnecessary restores/reads from expiry.php
OTOH, it could look hackish... Comments welcome.

bdunogier:
Updating the context before forking sounds more right, imho. Since every child will only publish once and disappear after that, it doesn't make sense to update in every child.

joaoinacio:
Updated.
Moved to main process just before publish, so it should run at most once every sleepInterval.

andrerom:
review ping @yannickroger @bdunogier

yannickroger:
+0,5 then. @bdunogier 's feedback is the most valuable on this one ;)

kernel/private/classes/ezpcontentpublishingqueueprocessor.php
```diff
@@ -353,6 +374,12 @@ private function getNextItem()
      */
     private $cleanupLastTime;
 
+     /**
+     * Time counter for refresh of cache/state information
+     * @var int
+     */
+    private $refreshLastTime;
```
bdunogier: I'd rather call it `lastRefreshTimestamp`. Not important, but... #naming

bdunogier:
The naming nitpick aside, +1.

joaoinacio:
@bdunogier: thanks. updated, should be ready for merge now.

yannickroger:
@joaoinacio you can merge it then ;)

georgfranz:
Hi,

I think the patch is a good approach and will solve the most issues.

But it won't solve the genuin problem:

If you have a large site with a lot of editors, async enabled, you have a lot of processes running at the same time. Some are faster, some are slower. There are fast publishing processes, some slower (with lot of workflow events attached) - and some admin is moving a tree by command line etc.

The main problem is:
The expiry.php is only fetched once by 
$expiryHandler = eZExpiryHandler::instance();

It will be never refreshed within one process. 

Example:
process a) 1 editor is moving a tree from A to B.
process b) 1 editor is editing the user roles (user cache)

Both processes are started at the near same time and are reading the same expiry.php.

So process a) is lasting longer as process b). Process b) is writing the user-cache key which is updated and process a) is writing the view cache key and the template block key.

The result is: Process a) is overwriting the changes of process b) (in that case the user info cache key)

Example from ezcache.php class:

static function clearTemplateBlockCache( $cacheItem )
    {
        eZExpiryHandler::registerShutdownFunction();
# this getting the old, outdated expiry.php
        $expiryHandler = eZExpiryHandler::instance();
        $expiryHandler->setTimestamp( 'global-template-block-cache', time() );
# this updates the 'global-template-block-cache' but overrides e.g. the content-view-cache key with an old value
        $expiryHandler->store();
    }

I would solve this problem this way:

Change this method of eZExpiryHandler:

/**
     * Stores the current timestamps values to cache
     */
    function store()
    {
        if ( $this->IsModified )
        {
            $this->CacheFile->storeContents( "<?php\n\$Timestamps = " . var_export( $this->Timestamps, true ) . ";\n?>", 'expirycache', false, true );
            $this->IsModified = false;
        }
    }

1. Reload the expiry.php
2. Do a safety check if the keys which should be updated are older than the current one. If they have the same or a newer timestamp, don't update (e.g. the current process is running longer and two other have already updated the expiry.php meanwhile)
3. Update only the keys which were changed within the current process (don't update all keys)

What do you think?

Best wishes,
Georg.

joaoinacio:
Hi Georg,

The points you mention do seem valid, though this is a more "generic" problem of concurrent expiry handling (not necessarily related to Async Publishing).
The main issue is loading/storing could be an expensive operation under (DFS) cluster, so updating (read + writing) the file could potentially cause some big performance hits... (f.e. due to locking latency, etc)
In addition, AFAICT, for cluster mode there is always a small chance for the expiry keys to be overwritten by other node (there is already locking but not at the key level)

Anyway perhaps it would be best to open a separate JIRA issue for this?

georgfranz:
Yeah, I will open a separate issue. I would be happy if we solve the async issue first in the suggested way.

georgfranz:
https://jira.ez.no/browse/EZP-23908

glye:
PR for EZP-23908: https://github.com/ezsystems/ezpublish-legacy/pull/1139

PR #779, opened 2013-10-04
Fix EZP-21686 - ezinfo/copyright has extra '\' that don't allow the links to work

pedroresende:

PR #1134, opened 2015-01-13
Fix EZP-21686: ezinfo/copyright has extra "\" that don't allow the links...

glye:
... to work

Alternative solution to PR #779. Implemented with nowdoc segment rather than quoted strings. Updated year to 2015. Added mailto-link for email address.

https://jira.ez.no/browse/EZP-21686

### DIFF
--- a/kernel/ezinfo/copyright.php
+++ b/kernel/ezinfo/copyright.php
@@ -8,34 +8,33 @@
 
 $Module = $Params['Module'];
 
-$text =
-## BEGIN COPYRIGHT INFO ##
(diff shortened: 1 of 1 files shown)


glye:
Sorry, forgot to check the links, they have rotted. Will need to find out were they are now, and push an update.

glye:
Have fixed rotted links to licence and contact page, and updated licence name and version number.

yannickroger:
+1

andrerom:
+1

PR #1135, opened 2015-01-13
Fix EZP-23879: ezinfo shows wrong copyright info after move to composer

glye:
Several changes:
- Removed hardcoded license text, instead show the whole contents of LICENSE (in legacy root) in an iframe
- Copyright year is updated to 2015
- The "What is eZ Publish?" text was split in two paragraphs in the source, this was invisible in html since there was no tagging. These are added now.
- Several text fixes (speling, punctuation etc)
- "eZ Systems", "eZ Publish" in the license text were converted to links, this is not done anymore since A) it is redundant, and B) we are now loading the whole LICENSE file and we should not mess with that - lawyers are involved.

https://jira.ez.no/browse/EZP-23879

### DIFF
--- a/design/admin/templates/ezinfo/about.tpl
+++ b/design/admin/templates/ezinfo/about.tpl
@@ -13,12 +13,14 @@
 
 <h3>{'What is eZ Publish?'|i18n( 'design/admin/ezinfo/about' )}</h3>
 {if is_set( $what_is_ez_publish )}
-    <p>{$what_is_ez_publish}</p>
+    {$what_is_ez_publish}
 {/if}
 
 <h3>{'License'|i18n( 'design/admin/ezinfo/about' )}</h3>
-{if is_set( $license )}
-    <p>{$license}</p>
+{if $license}
+    <pre style="height: 300px; overflow:scroll;">{$license|wash}</pre>
+{else}
+    <p>{'Could not load LICENSE file! You should have a LICENSE file in your eZ Publish root directory.'|i18n( 'design/admin/ezinfo/about' )}</p>
 {/if}
 
 {if and( is_set( $contributors ), is_array( $contributors ), count( $contributors )|ge( 1 ) )}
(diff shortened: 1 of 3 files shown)


glye:
NB:
- extension/ezdemo/design/ezdemo/templates/ezinfo/about.tpl will need the same modification as the others, though not possible in this PR
- We could remove the copyright year and instead refer to the file displayed above it...?
- The iframe is hardcoded to 600x300 CSS pixels, can we live with that?

design/admin/templates/ezinfo/about.tpl
```diff
@@ -13,12 +13,14 @@
...
 {if is_set( $what_is_ez_publish )}
-    <p>{$what_is_ez_publish}</p>
+    {$what_is_ez_publish}
 {/if}
 
 <h3>{'License'|i18n( 'design/admin/ezinfo/about' )}</h3>
-{if is_set( $license )}
-    <p>{$license}</p>
+{if $license}
+    <iframe width="600" height="300" sandbox srcdoc="<html><body>{$license}</body></html>"></iframe>
```
andrerom: hmm.. srcdoc doesnt seem to be supported on older browsers and IE, @dpobel any alternative suggestions? Could be pre tag or something, would be more fitting maybe.

andrerom:
> We could remove the copyright year and instead refer to the file displayed above it...?

yes please, it will get correct value in the content from the LICENSE file anyway now.

glye:
Notice it's loading LICENSE from legacy root, not main root - are we OK with that? Should always be the same, right...?

glye:
Removed copyright year, referring to License file now:
"Copyright for eZ Publish is included in the License shown above. Portions are copyright by other parties. A complete list of all contributors and third-party software follows."

design/admin/templates/ezinfo/about.tpl
glye: Right, looks like pre overflow:auto should work, then we also avoid the hackish html formatting.

glye:
Changed iframe to scrolling pre, html formatting removed.

andrerom:
> Notice it's loading LICENSE from legacy root, not main root - are we OK with that?

yes, usually should be the same. And safer from legacy root as it is the one in package which we have possibility to update (so will reflect new copyright year on sub releases).

andrerom:
+1

yannickroger:
+1

glye:
PR for the same fix in ezdemo: https://github.com/ezsystems/ezdemo/pull/39

PR #1136, opened 2015-01-13
Fix EZP-23890: better sort index elements in db schema def. extraction

gggeek:


andrerom:
hmm.. I wasn't expecting the change to also go here. any possible regressions this might cause?

gggeek:
q: "here" as in kernel-vs-oracle-extension or as opposed to something else?

I could possibly have applied the change at a higher level (e.g. the code doing schema serialization / dumping), but it feels more correct architecturally-wise to do it here.

I do not expect any regressions, because:
- the keys of the array (positions of index elements) are still mapped to the values (elements)
- the ez code which deals with going from schema definition to sql, and to check for differences in schema definitions already treats that array as a hash (or index rebuilding would never have been working)
- any other code which currently treats that array as positional instead of hash is currently broken and will in fact get fixed by this change (but since I know of no reported bugs with this, and its been a long time this code has been around, I suspect that there is no such thing out in the wild)

But just to be 200% sure, I can run a couple more test about the diff thing, by manually altering the order of index elements in a live db, and checking what is reported before and after the patch

gggeek:
Mmm, hold the horses!
It looks like the existing code which rebuilds the sql out of the schema does NOT treat that index definition as a hash, bur rather as a positional  array.

I definitely need to do more testing, but either I am mistaken in my understanding of the code, or there is a quite nasty bug lurking deep in here...

gggeek:
For the lolz: the extraction of index elements into a hash structure was written by none other than Derick, in January 2004

For the record: I am not 100% sure but iirc this, if confirmed, is not the 1st confusion in the schema file between positional arrays and hashes, one ore two more were already found and fixed in the past (but issues have to be dug out). The fact that the schema format was never documented but just dumped is a good candidate root cause.

andrerom:
which is exactly why I'm not a big fan of changes in the logic ;)

gggeek:
A bit more exploration led me to:

- oracle schema code: definitely a bug. Current code generates unsorted hashes for index elements, but import code treats it as a positional array. The PR to fix that is there should imho go in

- postgres schema code: the sql to extract index definition is more complex there, and the php code actually already takes into account that it has to build a positional array, starting at element 0 (line 298 in ezpgsqlschema.php)

- mysql schema code: here php creates a hash, and relies on a SHOW INDEX sql call. It does create a hash with keys which correspond to the returned Seq_in_index field, but it seems that the mysql devs were kind enough to always return index elements properly sorted. This is not part of the docs as far as I can see (http://dev.mysql.com/doc/refman/5.0/en/show-index.html), but we can assume it never failed so far (since no bug was reported in all these years)

So I'm fine with closing this PR, and merging the ezoracle one

andrerom:
POV @bdunogier ?

bdunogier:
If we can fix this without touching legacy, it's honestly better.

andrerom:
ok, yeah we agreed to skip this.

PR #1139, opened 2015-01-20
Fix EZP-23908: expiry.php race condition

glye:
Restore timestamps before saving, to reduce chance of race condition issues

This is the easy solution. The complete solution would be an asyncronous process with expiry request stored in db and a cron daemon which processes them at short intervals. I consider that overkill for legacy, which should not have new development.

Unknown: How much does this fix affect performance in cluster environments?

https://jira.ez.no/browse/EZP-23908

### DIFF
--- a/lib/ezutils/classes/ezexpiryhandler.php
+++ b/lib/ezutils/classes/ezexpiryhandler.php
@@ -61,6 +61,24 @@ class eZExpiryHandler
      */
     function store()
     {
+        if ( !$this->IsModified )
+        {
+            return;
+        }
+
+        // EZP-23908: Restore timestamps before saving, to reduce chance of race condition issues
+        $modifiedTimestamps = $this->Timestamps;
+        $this->restore();
+
+        // Apply timestamps that have been added or modified in this process
+        foreach ( $modifiedTimestamps as $name => $value )
+        {
+            if ( $value > self::getTimestamp( $name, 0 ) )
+            {
+                $this->setTimestamp( $name, $value );
+            }
+        }
+
         if ( $this->IsModified )
         {
             $this->CacheFile->storeContents( "<?php\n\$Timestamps = " . var_export( $this->Timestamps, true ) . ";\n?>", 'expirycache', false, true );


lib/ezutils/classes/ezexpiryhandler.php
```diff
@@ -61,6 +61,26 @@ static function fetchData( $path )
...
+        $this->restore();
+
+        // Apply timestamps that have been added or modified in this process
+        foreach ( $modifiedTimestamps as $name => $value )
+        {
+            if ( !array_key_exists( $name, $this->Timestamps ) ||
+                 $modifiedTimestamps[$name] > $this->Timestamps[$name] )
+            {
+                $this->Timestamps[$name] = $modifiedTimestamps[$name];
+                $this->IsModified = true;
```
joaoinacio: Nitpick: IMHO, something like the following would look a bit cleaner:
```
if ( $value > $this->getTimestamp( $name, 0 )
{
     $this->setTimestamp( $name, $value )
}
```

glye:
Re: Travis, I don't see how expiry.php can cause eZImageTypeRegression::tearDown() to fail. Bogus? Then again, messing with expiry can affect a lot of things...

lib/ezutils/classes/ezexpiryhandler.php
glye: Fair enough, except that getTimestamp() is static :)

joaoinacio: Doh indeed, good catch! 
Well something like ```! $this->hasTimestamp( $name ) || $value > $this->timestamp($name)``` ?

joaoinacio:
No idea about the previous travis fail either, very weird glitch...

Regarding the performance aspect, a possible improvement would be to lazy-load expiry data on the ```hasTimstamp()``` / ```getTimestamp()``` / ```timestamp()``` methods.
In theory, this could potentially reduce the fetch/restore by 1 in some cases?

In any case, this should pretty much resolve the issue - even if it's not 100% fail-safe (no additional locking, etc) I don't see any other solution that wouldn't involve a major rewrite/implementation...
So FYIW, +1

glye:
@joaoinacio Thanks. In what cases would lazy loading help?

joaoinacio:
@glye:
I imagine there are a few code paths were expiry data isn't used except for actually expiring content on shutdown?
In practice I have no idea what the percentage of the usage/requests actually needs expiry data, but lazy-loading would also have the potential benefit of reading "fresher" values...

glye:
@georgfranz Hi! This is not needed anymore, after the change suggested by @joaoinacio. isModified is set to false by the restore() call before the foreach, and set to true by setTimestamp().

georgfranz:
ah, ok! tx for explanation!

lib/ezutils/classes/ezexpiryhandler.php
```diff
@@ -61,6 +61,24 @@ static function fetchData( $path )
      */
     function store()
     {
+        if ( !($this->IsModified) )
```
lolautruche: Inner parenthesis are not needed

lolautruche:
+1

PR #1140, opened 2015-01-28
Fix EZP-23725: Wrong path_identification_string with custom RootNode

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-23725

When a different RootNode is specified for a SA, along with a PathPrefix, it's generated path will be empty;
However path_string should remain the full value.

Because path_string is based on each node's parent, simply not updating the new RootNode to empty should resolve the issue.

### DIFF
--- a/kernel/classes/ezcontentobjecttreenode.php
+++ b/kernel/classes/ezcontentobjecttreenode.php
@@ -3613,7 +3613,11 @@ class eZContentObjectTreeNode extends eZPersistentObject
         {
         }
 
-        $this->updatePathIdentificationString( $pathIdentificationName );
+        // Only update pathIdentificationString if new name is defined (not set for RootNode)
+        if ( $pathIdentificationName )
+        {
+            $this->updatePathIdentificationString( $pathIdentificationName );
+        }
 
         $languageID = $obj->attribute( 'initial_language_id' );
         $cleanup    = false;


andrerom:
this does not fix the full issue does it? it will still write wrong path identification to db when the variable has value (so sub node of root node).

And what if you move a node to become root node and RootNode=2, should it not be updated?

joaoinacio:
child tree of the new RootNode get the path string calculated from their parent, so it should remain consistent up to real root ('').

Im not sure if I can see a clear use case for that, but perhaps it should be set if empty only for 2, or if the parent is 1?

yannickroger:
ping @andrerom :)

andrerom:
@yannickroger I still don't understand why this fixes the issue so can not help you before we clear that out.

crevillo:
I don't understand it either. I add that looks a bit weird to me that we're entering in the [function](https://github.com/joaoinacio/ezpublish-legacy/blob/EZP-23725_RootNode_PathIdentString/kernel/classes/ezcontentobjecttreenode.php#L3505), we do all those things and we don't call updatePathIdentificationString for that specific case... 
In other words, we would need to update path and not update path string? 
Saying that without looking deeply to the code, i admit.

joaoinacio:
I'll try to summarize it:
* Normally, RootNode is 2 (home), and it's pathString is ''
* Siteaccess A is configured for rootNode 100 (/home/A)
* When editing content through frontend SA A, the pathIdentString will be updated and will start from rootNode 100 (empty), for example, but old nodes will remain as before:
```
|  /1/2/100/ |  |
|  /1/2/100/200 | TestFolder |
|  /1/2/100/210 | A/OldFolder |
```
This happens because since it the RootNode is 100, it the pathstring will be considered from there.
All new/updated children will be updated to use it's parent, but old nodes will remain with old pathString.

If a new Siteaccess B is introduced, it can generate even more conflicts with the pathIdentificationString (duplicates):
```
|  /1/2/100/ |  |
|  /1/2/101/ |  |
|  /1/2/100/200 | TestFolder |
|  /1/2/101/201 | TestFolder |
```

AFAICT, not saving empty pathIdentStrings resolves the problem (for rootnodes and all their children) as it will keep consistency:

```
|  /1/2/100/ | A |
|  /1/2/101/ | B |
|  /1/2/100/200 | A/TestFolder |
|  /1/2/100/210 | A/OldFolder |
|  /1/2/101/201 | B/TestFolder |
```

Hope this clarifies the problem/fix a bit

andrerom:
Ok, I get it now.

However the fix still looks like a workaround basically which is why I questioned it in the first place.
Ideally I think we should have introduced a new setting called SiteNode, which is what we should inject instead, use it only in front end cases, and reflect that in doc.

Meaning RootNode is kept to mean only one thing, the root of the content structure in the repository.

yannickroger:
The main advantage of the setting is BC. So we can backport and distribute the fix (composer) without worrying too much.

joaoinacio:
@andrerom: It seems to me that introducing the new setting could result in a huge amount of work with testing, and even possible regressions?
There are at the very least tens of uses of RootNode spread throughout legacy in both code and templates, some of which not fully (or at all) documented...
IMHO, given the plans for legacy, I am not sure if this work is worth it, tbh...

andrerom:
well, changing view layer vs changing lower layers like this basically.
Recommendations for multisite stated certain places that you should use RootNodeDepth some places, which is also just a view feature, but a flawed one as it breaks down on most other modules then content.

yannickroger:
ping @andrerom : Did you have a chance to provide a gist to @joaoinacio with more info (as suggested yesterday) ?

andrerom:
no, its in my todo :)

andrerom:
Look into it now, and I see what @joaoinacio meant that there is tons of possible issues with introducing new setting also.

So I change to +1 on this, better than nothing.

yannickroger:
+1. Let's have it QA only on master before backporting. Ready for merge @joaoinacio

andrerom:
@yannickroger better to backport and provide to QA so we don't have to backport as separate issue.

pspanja:
After staring at the code for some time the fix looks ok to, also I don't see that URL aliases will be affected in any way by this change. So it's a +1 from me.

yannickroger:
@joaoinacio merge ?

PR #1142, opened 2015-02-19
Fix EZP-24030: Relationlist contains wrong content after content removal

glye:
Update the attribute object relations xml whenever the main node id of an object changes. Always set it to main node id.

https://jira.ez.no/browse/EZP-24030

### DIFF
--- a/kernel/classes/datatypes/ezobjectrelationlist/ezobjectrelationlisttype.php
+++ b/kernel/classes/datatypes/ezobjectrelationlist/ezobjectrelationlisttype.php
@@ -1878,6 +1878,38 @@ class eZObjectRelationListType extends eZDataType
         return $return;
     }
 
+    /*!
+     Replaces node ID for object with given ID in the relations list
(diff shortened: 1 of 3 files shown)


andrerom:
Be aware that some time ago doc was updated to reflect that these kind of properties are only updated on re publish of the origin object.

some mentions of this at least here, might also be other places:
https://doc.ez.no/eZ-Publish/Technical-manual/4.x/Reference/Datatypes/Object-relations

glye:
@andrerom Right. But this particular inconsistency is not fixed on republishing.

andrerom:
hmm, that is something else.

glye:
Well, in this case the bug is not triggered by modifying the relation, but rater by switching main node. So we can refer to "at the time when the relation was modified" in the doc, close as invalid, and give the patch to the original reporter as a custom workaround, if they really must have it.

andrerom:
ok

PR #1143, opened 2015-02-19
Fix EZP-23928: Preview cache not being updated

lolautruche:
> https://jira.ez.no/browse/EZP-23928

When saving a draft (not publishing), preview won't reflect the changes. A previously cached content is displayed instead.
Only applies if the frontend siteaccess is *not* using `legacy_mode`.

This patch introduces a new `content/cache/version` event.


andrerom:
+1

yannickroger:
+1

lolautruche:
That was quick :smile:

PR #1145, opened 2015-02-27
EZP-23725: Add SiteNode support and recommend over RootNode for multisit...

andrerom:
...e setup

Other PR's:
- [x] ezdemo https://github.com/ezsystems/ezdemo/pull/41
- [ ] ezwebin/ezflow


andrerom:
On further examination this will also have tons of side effects on redirection behavior during front end actions (edit, add to basket, ...), so closing in favour of original proposal.

PR #1146, opened 2015-03-03
EZP-24099: Generate relationlist metadata for Search (Solr) using published content

andrerom:
Using current (published) content instead of original data (from when relation was added).

See change for further info on why, but in short this is already the case for generation of title and to/from string and online documentation (that code should never expect meta data like version, node id and so on to be up-to-date).

Credits: Original patch from @phalasz can be found in https://jira.ez.no/browse/EZP-24099
         Expanded in this version with more doc, cs, and cleanup of variable names for readability.

### DIFF
--- a/kernel/classes/datatypes/ezobjectrelationlist/ezobjectrelationlisttype.php
+++ b/kernel/classes/datatypes/ezobjectrelationlist/ezobjectrelationlisttype.php
@@ -814,18 +814,33 @@ class eZObjectRelationListType extends eZDataType
         return $doc;
     }
 
+    /**
+     * Returns xml identifier to attribute map
+     *
+     * Key is xml identifier as used in attribute data_string format, and value is the name exposed in attribute.
+     *
+     * Warning: Besides contentobject_id most attributes are not updated on changes and hence represent the values when
+     *          the relation was first created. These should ideally have been prefixed with "original" (see inline).
+     *          You can see this in affect all over this datatype that for instance version number from attribute
+     *          source is newer used for basis to generate things like title, metadata or toString value.
+     *
+     * @return array
+     */
     static function contentObjectArrayXMLMap()
     {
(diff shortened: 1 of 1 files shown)


kernel/classes/datatypes/ezobjectrelationlist/ezobjectrelationlisttype.php
```diff
@@ -1564,15 +1579,21 @@ function metaData( $contentObjectAttribute )
...
             else
             {
-                $subObjectVersion = $relationItem['contentobject_version'];
-                $object = eZContentObject::fetch( $subObjectID );
+                $subObjectVersionNum = $relationItem['contentobject_version'];
+                $subObject = eZContentObject::fetch( $subObjectID );
+
+                // Using last version of object (version inside xml data is the original version)
+                $subCurrentVersionObject = $subObject->currentVersion();
+                if( $subCurrentVersionObject instanceof eZContentObjectVersion ) {
```
crevillo: nitpick: space after the if  bracket in new line (even dunno if we are more relaxed with this in legacy...)

kernel/classes/datatypes/ezobjectrelationlist/ezobjectrelationlisttype.php
```diff
@@ -1564,15 +1579,21 @@ function metaData( $contentObjectAttribute )
...
+                    $subObjectVersionNum = $subCurrentVersionObject->attribute('version');
```
crevillo: cs: spaces before and after ```'version'```

andrerom:
Thanks for being awake @crevillo, fixed :)

bdunogier:
+1

andrerom:
review ping @glye / @pspanja

pspanja:
+1

PR #1147, opened 2015-03-04
Update ezautolinkoperator.php

tompsonx:
Add support for text which is started from www. without protocol.
For example change text "www.ez.no" to link tags.

### DIFF
--- a/kernel/common/ezautolinkoperator.php
+++ b/kernel/common/ezautolinkoperator.php
@@ -33,24 +33,35 @@ class eZAutoLinkOperator
                                             'default' => null ) );
     }
 
-    function formatUri( $url, $max )
+    static function formatUri( $url, $max, $method = '' )
     {
-        $text = $url;
+        if ( $method == 'www.' )
+        {
+            $textFull = $method . $url;
+            $text = $textFull;
+            $url = 'http://' . $method . $url;
+        }
+        else
+        {
+            $textFull = $url;
+            $text = $textFull;
+            $url = $method . $url;
+        }
         if (strlen($text) > $max)
         {
             $text = substr($text, 0, ($max / 2) - 3). '...'. substr($text, strlen($text) - ($max / 2));
         }
-        return "<a href=\"$url\" title=\"$url\">$text</a>";
+        return "<a href=\"$url\" title=\"$textFull\">$text</a>";
     }
 
     /*!
      \static
     */
-    function addURILinks( $text, $max, $methods = 'http|https|ftp' )
+    static function addURILinks( $text, $max, $methods = 'http|https|ftp' )
     {
         return preg_replace(
-            "`(?<!href=\"|href='|src=\"|src='|value=\"|value=')($methods):\/\/[\w]+(.[\w]+)([\w\-\.,@?^=%&:\/~\+#;*\(\)\!]*[\w\-\@?^=%&\/~\+#;*\(\)\!])?`e",
-            'eZAutoLinkOperator::formatUri("$0", '. $max. ')',
+            "`(?<!href=\"|href='|src=\"|src='|value=\"|value=')((?:(?:$methods):\/\/)|www.)([\w]+(?:.[\w]+)(?:[\w\-\.,@?^=%&:\/~\+#;*\(\)\!]*[\w\-\@?^=%&\/~\+#;*\(\)\!])?)`e",
+            'eZAutoLinkOperator::formatUri("$2", '. $max. ', "$1")',
             $text
         );
     }


brookinsconsulting:
-1 without ini setting disabling this feature by default

tompsonx:
I added ini setting disabling this feature by default.

brookinsconsulting:
Hello @tompsonx

Thank you for you prompt response.

We have some additional comments but first would you create a jira.ez.no issue ticket and share the link here in the pr?

kernel/common/ezautolinkoperator.php
```diff
@@ -42,17 +42,48 @@ function formatUri( $url, $max )
         }
         return "<a href=\"$url\" title=\"$url\">$text</a>";
     }
+    
+    static function formatUriSimple( $url, $max, $method = '' )
```
brookinsconsulting: Do we really need a separate function? We really don't think so. We also do not care for the current name as vague. We think you features would be better placed within the original function instead.

kernel/common/ezautolinkoperator.php
```diff
@@ -60,6 +91,7 @@ function modify( $tpl, $operatorName, $operatorParameters, $rootNamespace, $curr
     {
         $ini = $tpl->ini();
         $max = $ini->variable( 'AutoLinkOperator', 'MaxCharacters' );
+        $simpleWww = ( $ini->variable( 'AutoLinkOperator', 'ChangeUrlWithoutProtocol' ) == 'enabled' );
```
brookinsconsulting: We think that this var name should be more clear and specific on what it represents. While perhaps too long even $linkWwwUrlWithoutProtocol or $linkWwwUrl would be better

settings/template.ini
```diff
@@ -72,6 +72,10 @@ Methods[]=ftps
 Methods[]=webdav
 Methods[]=webdavs
 
+# Add support for text which is started from www. without protocol.
+# For example change text "www.ez.no" to link tags.
+ChangeUrlWithoutProtocol=disabled
```
brookinsconsulting: We think that this setting name should be more clear and specific on what it represents. LinkWwwUrlsWithoutProtocol would be better because it is very clear and describes exactly what your feature does. Your current setting name is too vague (Re: use of terms Change and Url (singular)) and technically inaccurate as it only links urls starting with 'www.'.

tompsonx:
Thank you for your constructive comments.
I rebuilt a little code. What do you think about such modifications.
I added static properties, changed preg_replace to preg_replace_callback and added parameter for remove protocols from text.

I create jira issue: EZP-24114.

brookinsconsulting:
Link to issue ticket, https://jira.ez.no/browse/EZP-24114

kernel/common/ezautolinkoperator.php
```diff
@@ -10,6 +10,13 @@
 
 class eZAutoLinkOperator
 {
+    static $max = 72;
+    
+    static $isLinkWwwUrl = false;
```
brookinsconsulting: This variable name is not so good, as it suggests by it's name (to us at any level) that the link in question is a www. style link but in truth the variable is only a boolean indicating that the operator use link www. type text. For these reasons we think this variable name should be changed. `$linkWwwUrls` (plural) is in our opinion a much better name. We request you change this variable name.

kernel/common/ezautolinkoperator.php
```diff
@@ -10,6 +10,13 @@
...
+    
+    static $hideProtoclText = '';
```
brookinsconsulting: This variable name should also be changed to `$hideProtocolsText` at the very least. We think this was merely a spelling typo. We request you change this variable name.

kernel/common/ezautolinkoperator.php
```diff
@@ -33,7 +40,12 @@ function namedParameterList()
                                             'default' => null ) );
     }
 
-    function formatUri( $url, $max )
+    /*!
+     Leave for backward compatibilty.
+     \static
+     \deprecated
+    */
```
brookinsconsulting: For all class methods you touch / change / modify (at least) in this class file please convert the function doxygen comments to phpdoc (the code standard for all new eZ Publish Legacy code). For all new methods please create new phpdoc function comments. This step is required.

kernel/common/ezautolinkoperator.php
```diff
@@ -42,15 +54,52 @@ function formatUri( $url, $max )
         }
         return "<a href=\"$url\" title=\"$url\">$text</a>";
     }
+    
+    static function formatUriCallback( $matches )
```
brookinsconsulting: This method needs a phpdoc comment block above it

kernel/common/ezautolinkoperator.php
```diff
@@ -42,15 +54,52 @@ function formatUri( $url, $max )
...
+    {
+        $method = $matches[1];
+        $www = $matches[2];
```
brookinsconsulting: This variable name should also be changed to `$hostname` at the very least. We request you change this variable name.

kernel/common/ezautolinkoperator.php
```diff
@@ -42,15 +54,52 @@ function formatUri( $url, $max )
...
+        $domainPath = $matches[3];
```
brookinsconsulting: This variable name should also be changed at the very least. We suggest `$domainAndUri` as a more clear and verbose replacement. We request you change this variable name.

kernel/common/ezautolinkoperator.php
```diff
@@ -42,15 +54,52 @@ function formatUri( $url, $max )
...
+        
+        $url = $method . $www . $domainPath;
+        
+        if ( self::$isLinkWwwUrl && $www == 'www.' )
+        {
+            $url = 'http://' . $www . $domainPath;
+        }
+        
+        $textFull = $url;
```
brookinsconsulting: We think this variable name should also be changed. We suggest `$urlLinkText` as a more clear and verbose replacement. We request you change this variable name.

kernel/common/ezautolinkoperator.php
```diff
@@ -42,15 +54,52 @@ function formatUri( $url, $max )
...
+        
+        if ( self::$hideProtoclText != '' )
+        {
+            $textFull = preg_replace( '`^(?:' . self::$hideProtoclText . '):\/\/(.*)$`', '$1', $textFull );
+        }
+        
+        $text = $textFull;
```
brookinsconsulting: Please add a blank line here for increased readability.

kernel/common/ezautolinkoperator.php
brookinsconsulting: This class method name still leaves something to be desired. We do not think that 'Callback' method name text is clear or needed. We suggest `formatUrlLinkHtml` as a replacement for this class method name.

kernel/common/ezautolinkoperator.php
```diff
@@ -42,15 +54,52 @@ function formatUri( $url, $max )
...
+        if (strlen($text) > self::$max)
+        {
+            $text = substr($text, 0, (self::$max / 2) - 3). '...'. substr($text, strlen($text) - (self::$max / 2));
+        }
```
brookinsconsulting: Please add a blank line here for increased readability.

kernel/common/ezautolinkoperator.php
```diff
@@ -42,15 +54,52 @@ function formatUri( $url, $max )
...
+      $max - leave for backward compatibilty
      \static
     */
-    function addURILinks( $text, $max, $methods = 'http|https|ftp' )
+    static function addURILinks( $text, $max = false, $methods = 'http|https|ftp' )
     {
-        return preg_replace(
-            "`(?<!href=\"|href='|src=\"|src='|value=\"|value=')($methods):\/\/[\w]+(.[\w]+)([\w\-\.,@?^=%&:\/~\+#;*\(\)\!]*[\w\-\@?^=%&\/~\+#;*\(\)\!])?`e",
-            'eZAutoLinkOperator::formatUri("$0", '. $max. ')',
+        if ( $max !== false && is_int( $max ) )
```
brookinsconsulting: Please use if conditional statement brackets here to conform to the coding standard and increase readability.

kernel/common/ezautolinkoperator.php
```diff
@@ -42,15 +54,52 @@ function formatUri( $url, $max )
...
+            self::$max = $max;
+        
+        if ( self::$isLinkWwwUrl )
```
brookinsconsulting: Please use if conditional statement brackets here to conform to the coding standard and increase readability.

kernel/common/ezautolinkoperator.php
```diff
@@ -42,15 +54,52 @@ function formatUri( $url, $max )
...
+            $mainSearchText = "(?:((?:$methods):\/\/)|(www\.))";
```
brookinsconsulting: This variable name is not so good, it's simply just not very clear. We think that `$searchExpPart`is a (somewhat) more accurate name (at least). We request you change this variable name.

kernel/common/ezautolinkoperator.php
```diff
@@ -59,20 +108,25 @@ function addURILinks( $text, $max, $methods = 'http|https|ftp' )
     function modify( $tpl, $operatorName, $operatorParameters, $rootNamespace, $currentNamespace, &$operatorValue, $namedParameters, $placement )
     {
         $ini = $tpl->ini();
-        $max = $ini->variable( 'AutoLinkOperator', 'MaxCharacters' );
+        self::$max = $ini->variable( 'AutoLinkOperator', 'MaxCharacters' );
```
brookinsconsulting: Please add a blank line here for increased readability.

kernel/common/ezautolinkoperator.php
```diff
@@ -59,20 +108,25 @@ function addURILinks( $text, $max, $methods = 'http|https|ftp' )
...
-            $max = $namedParameters['max_chars'];
+            self::$max = $namedParameters['max_chars'];
         }
 
         $methods = $ini->variable( 'AutoLinkOperator', 'Methods' );
         $methodText = implode( '|', $methods );
 
+        self::$isLinkWwwUrl = ( $ini->variable( 'AutoLinkOperator', 'LinkWwwUrlsWithoutProtocol' ) == 'enabled' );
+        
+        $hideProtocls = $ini->variable( 'AutoLinkOperator', 'HideProtocols' );
```
brookinsconsulting: This variable name should also be changed to `$stripProtocolsText` at the very least. We think this was merely a spelling typo. We request you change this variable name.

settings/template.ini
```diff
@@ -72,6 +72,13 @@ Methods[]=ftps
 Methods[]=webdav
 Methods[]=webdavs
 
+# Add support for text which is started from www. without protocol.
+# For example change text "www.ez.no" to link tags.
+LinkWwwUrlsWithoutProtocol=disabled
+
+HideProtocols[]
```
brookinsconsulting: This variable name is not so good, as it suggests by it's name (to us at any level) that the protocol text is hidden, which is not exactly true. Instead in truth, when this feature is used the protocols are striped from the generated link text. For these reasons we think this setting name should be changed. `StripLinkProtocols` is in our opinion a better name. We request you change this setting name.

settings/template.ini
```diff
@@ -72,6 +72,13 @@ Methods[]=ftps
...
+#HideProtocols[]=http
```
brookinsconsulting: Now for a more positive comment :) We like that you give a specific example of how to use this setting.

settings/template.ini
```diff
@@ -72,6 +72,13 @@ Methods[]=ftps
 Methods[]=webdav
 Methods[]=webdavs
 
+# Add support for text which is started from www. without protocol.
+# For example change text "www.ez.no" to link tags.
```
brookinsconsulting: We think this setting comment block text should be re-written and clarified. Here is a suggested replacement:

`# Link text urls starting with 'www.' without protocol when using autolink template operator
# For example link text string "www.ez.no". Disabled by default to retain BC.
# Use either enabled or disabled.`

settings/template.ini
brookinsconsulting: This setting block requires a detailed comment explaining it's use. Here is a suggested example:
```
# Strip protocols from link text when using autolink template operator
# For example link text string "http://www.ez.no". Disabled by default to retain BC.
# Add link protocols to enable this feature.```

brookinsconsulting:
Hello @tompsonx 

Apologies for so many individual refactoring requests / comments.

We are striving to ensure that if this pull request is merged, it is done only when the code is of the very best possible quality.

In eZ Publish Legacy we are striving to ensure BC, strict code standards and reliability. In general these days we are actually trying very hard to *not* make large or unnecessary changes to the code.

For these reasons and (specifically) because of your regex changes. I think for this to be considered for merging we would prefer you also provide extensive unit test coverage of this class covering all previous / existing and new use cases.

Functions that would need to be tested would be those you added and modified. The goal is to make sure they behave the way they are supposed to in each unique use case (variable input use cases, and by our mark, there are quite a few which will need to be written).

Unit test are using phpunit (you can start by looking at tutorials online). You can find documentation on how to run legacy unit test here: http://www.ezpedia.org/ez/testing_ez_publish_test_system

You can add new tests in: https://github.com/ezsystems/ezpublish-legacy/blob/master/tests/tests/kernel/common/ezautolinkoperator_test.php

Cheers,
Brookins Consulting

PR #1148, opened 2015-03-09
EZP-24115 front multisite rootnode

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-24115

When using a custom RootNode (multisite settings), editing the root node through  the frontend will cause issues with the UrlAlias (the homepage / site root displays wrong content).

The problem is that data in the ezurlalias table (as well as the path_ident_string field in content_tree) should  not be specific and relative to a given siteaccess, but for the "whole" site (as defined in the backoffice).

This reverts commit for EZP-23725, and introduces a new setting `ContentTreeRootNode` in `content.ini`.
This setting is used solely in eZContentObjectTreeNode, for the purpose of updating the subtree paths (urlalias as well as path_ident_string).

### DIFF
--- a/kernel/classes/ezcontentobjecttreenode.php
+++ b/kernel/classes/ezcontentobjecttreenode.php
@@ -3613,11 +3613,7 @@ class eZContentObjectTreeNode extends eZPersistentObject
         {
         }
 
-        // Only update pathIdentificationString if new name is defined (not set for RootNode)
-        if ( $pathIdentificationName )
-        {
-            $this->updatePathIdentificationString( $pathIdentificationName );
-        }
+        $this->updatePathIdentificationString( $pathIdentificationName );
 
         $languageID = $obj->attribute( 'initial_language_id' );
         $cleanup    = false;
--- a/kernel/classes/ezcontentobjecttreenode.php
+++ b/kernel/classes/ezcontentobjecttreenode.php
@@ -3518,7 +3518,7 @@ class eZContentObjectTreeNode extends eZPersistentObject
 
         // Only set name if current node is not the content root
         $ini = eZINI::instance( 'content.ini' );
-        $contentRootID = $ini->variable( 'NodeSettings', 'RootNode' );
+        $contentRootID = $ini->variable( 'NodeSettings', 'ContentTreeRootNode' );
         $obj           = $this->object();
         $alwaysMask    = ( $obj->attribute( 'language_mask' ) & 1 );
         $languages     = $obj->allLanguages();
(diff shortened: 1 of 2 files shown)


andrerom:
Have you gone over all use of RootNode to verify what else would need to moved to ContentTreeRootNode?

What I proposed in the other PR but stopped at half way was exactly the opposite, introduce new setting that has the meaning of "SiteRootNode" and adjust ezpublish-kernel to inject that instead. ref: https://github.com/ezsystems/ezdemo/pull/41

joaoinacio:
As mentioned in the description I believe the main issues are caused by any changes to the DB data in an unwated way, that can potentially break f.e. any other siteaccesses on the same database:
TreeNode's `path_identification_string`, UrlAlias.

As for the rest of the usages, I'm not certain if anything else needs to be modified, as the general behavior seems to be what is expected.
Legacy docs are somewhat scarse, but from the SF side the following still holds:
`This parameter sets the location id of the content root. The API is jailed within this root, and nothing above this level will be accessible by default.`

I will perform some additional testing and dig a bit through any usages, but unless something is causing bad DB data it should be kept as-is, IMHO.

andrerom:
Don't look at symfony docs, we made a mistake there in how we use RootNode, that is the root cause of these issues, not anything in legacy. Which is why I think the safest is probably to just inject RootNodeDepth instead in Symfony part, or create a new setting for the webin/demo site root node.

pedroresende:
@bdunogier, @yannickroger, @lolautruche ping

yannickroger:
Shouldn't we add a BC message to tell people that are not using `2` has a root node to update the new setting.

joaoinacio:
Regarding the BC that would be for `5.9` in master, right?

yannickroger:
I would say 2015.03 ? ping @andrerom

bdunogier:
What do we do with @andrerom's comment ? Do we want to try that way instead ?

joaoinacio:
The problem with andre's approach, from what I can tell, is that 
* Most occurrences of 'RootNode' would need to be changed to a new setting, and SF stack updated.
* There are uses in extensions (I haven't checked all tbh) that also need to be updated accordingly.
* Backporting, distribution, and BC could be a big issue, specially considering template code and custom overrides on existing installations.

From what I can tell, IMHO all of the above seems unnecessary as the end result should be (more or less) the same as this PR...

bdunogier:
Did you see the same consequences, @andrerom ?​

andrerom:
Yes, but I have covered them before.

tharkun:
This PR should be closed as the jira ticket is closed and the PR written by @lolautruche are merged.

andrerom:
thanks @tharkun

PR #1149, opened 2015-03-12
Fix ezoe : alone character "0" inside xml custom and native tag replaced by space

yannschepens:
This is a strange case, but this kind of thing append.

I tried to create a Jira ticket but I didn't have account.


brookinsconsulting:
Hello @yannschepens 

You need to register for a 'free' account first. Please register here: https://login.ez.no/register?return=http://share.ez.no/

Then when you have your account created (and verified? I forget if they still do that, in any case) next you can login to jira.ez.no using the same user name and password and create an issue ticket.

Then share the url to the issue ticket here in this PR. Also rename your PR Title to include the issue ticket number (Follow the standard format for this).

Thank you. I hope this helps!

Cheers,
Brookins Consulting

extension/ezoe/tests/ezoexmltext_regression.php
```diff
@@ -391,6 +391,38 @@ public function providerMixedCaseAttributes()
             )
         );
     }
+    
+    /**
+     * Alone "0" character replaced by space into inline custom tag
+     */
```
brookinsconsulting: When you have created a detailed jira.ez.no issue ticket please update this comment to follow the format examples provided by other tests like this one, https://github.com/yannschepens/ezpublish-legacy/blob/master/extension/ezoe/tests/ezoexmltext_regression.php#L168

Specifically include the issue number and link url to the issue number following the above format example.

extension/ezoe/ezxmltext/handlers/input/ezoexmlinput.php
```diff
@@ -1329,13 +1329,13 @@ function &inputTagXML( &$tag, $currentSectionLevel, $tdSectionLevel = null )
...
                 {
-                    if ( !$childTagText ) $childTagText = '&nbsp;';
+                    if ( ! strlen($childTagText) ) $childTagText = '&nbsp;';
                     $output .= '<' . self::$nativeCustomTags[ $name ] . $customAttributePart . $styleString .
                                '>' . $childTagText . '</' . self::$nativeCustomTags[ $name ] . '>';
                 }
                 else if ( $inline === true )
                 {
-                    if ( !$childTagText ) $childTagText = '&nbsp;';
+                    if ( ! strlen($childTagText) ) $childTagText = '&nbsp;';
```
andrerom: cs, should be no space in front of function with the legacy ez coding standard.

andrerom:
+1 with nitpicks and issues once someone is able to create one.

@brookinsconsulting Seems to be some issues on share.ez.no, so probably not possible to register right now :(

brookinsconsulting:
@andrerom OK. Since I can not see into those share issues myself would you be so kind to update this PR with a comment when you think those 'issues' with registration / share are resolved?

yannschepens:
Updated ;)

brookinsconsulting:
@yannschepens 

Your improvements are good but ... you did not fix the CS issues that @andrerom is requiring you make first.

Once you fix the CS issues, please remember to rebase and flatten your commits into one commit (Like you have done in the past / already before).

Thank you for your continued support!

Cheers,
Brookins Consulting

andrerom:
Seems to be fixed now.

review ping @yannickroger @bdunogier (I think this is good for 5.3 and 5.4 backport as well)

PR #1150, opened 2015-03-18
Fix EZP-24135: Output buffer gets cleared after legacy call

glye:
Dirty trick to see if we are in legacy context, do ob_end_clean only when we are. Seems to work.

https://jira.ez.no/browse/EZP-24135

### DIFF
--- a/kernel/classes/ezscript.php
+++ b/kernel/classes/ezscript.php
@@ -235,8 +235,11 @@ class eZScript
     */
     function initialize()
     {
-        if( ob_get_length() != 0 )
-            ob_end_clean();
+        if ( !ezpKernel::hasInstance() )
+        {
+            if ( ob_get_length() != 0 )
+                ob_end_clean();
+        }
 
         // Initialize text codec settings
         $this->updateTextCodecSettings();
--- a/kernel/private/classes/ezpkernel.php
+++ b/kernel/private/classes/ezpkernel.php
@@ -120,6 +120,16 @@ class ezpKernel implements ezpWebBasedKernelHandler
     }
 
     /**
+     * Checks if the kernel has already been instantiated.
+     *
+     * @return bool
+     */
+    public static function hasInstance()
+    {
+        return self::$instance !== null;
+    }
+
+    /**
      * Returns the current instance of ezpKernel.
      *
      * @throws LogicException if no instance of ezpKernel has been instantiated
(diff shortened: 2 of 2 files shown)


andrerom:
Hmm.. ( hackish.. ;) )

Did an attempt to see why it is there to begin with, some clues but no clear answer:
https://github.com/ezsystems/ezpublish-legacy/commit/17103efe90be66cab408f87c506cd6ca8609fd61
https://github.com/ezsystems/ezpublish-legacy/commit/16e1f5f89e34fa1eb91bef0190d4473c896ba8f3#diff-17591b11c0bdab82578dc1ec724410d5

ping @lolautruche

glye:
Hackish, yes :) I did try to just remove the call and run a few legacy scripts, didn't see any problems then, but by no means a complete test. The oldest commit mentions runcronjobs.php, that might be it - I can see someone doing this as a sort of precaution since we don't know in advance what scripts it will call. (Not that I know any cronjob scripts that depend on the ob being empty.)

kernel/classes/ezscript.php
```diff
@@ -235,8 +235,17 @@ function startup()
     */
     function initialize()
     {
-        if( ob_get_length() != 0 )
-            ob_end_clean();
+        try
+        {
+            ezpKernel::instance();
```
lolautruche: Why would this throw an exception?

glye: Hm, maybe it won't. I grabbed this from https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/private/classes/ezpcontentpublishingprocess.php#L174 without really understanding it. Do we have a test for what I'm trying to do here?

lolautruche: OK I get it. But this is really hackish...
Actually we should better have an instance of `ezpKernel` ready in CLI, like in platform stack

andrerom: I think you might have to expand on that @lolautruche

lolautruche: OK, I think the best would be to move `hasInstance()` method from [`\eZ\Publish\Core\MVC\Legacy\Kernel`](https://github.com/ezsystems/LegacyBridge/blob/master/mvc/Kernel.php#L71-L79) to `ezpKernel`. Then, instead of try/catch, you would do:

```php

```if ( !ezpKernel::hasInstance() )
{
    if ( ob_get_length() != 0 )
        ob_end_clean();
}

glye:
Moved hasInstance() method from \eZ\Publish\Core\MVC\Legacy\Kernel to ezpKernel
Use hasInstance() in eZScript
Use hasInstance() in Kernel::instance()
CS fix in Kernel::instance()

LegacyBridge PR coming up soon...

lolautruche:
+1

andrerom:
+1

PR #1151, opened 2015-03-23
Fix EZP-24154: Overriding template with section_identifier fails when cached

andrerom:
Issue: https://jira.ez.no/browse/EZP-24154

`section_identifier` is not cached and not re applied after loading from cache, fixed in this PR by adding to `content_info`.


andrerom:
review ping @glye @yannickroger

glye:
Must be a very old bug...? +1

andrerom:
yes, since 4.4 I think

yannickroger:
+1

andrerom:
Thanks, merging :)

PR #1152, opened 2015-03-26
Fix EZP-24100: Image content with big multibyte charset names won't be correctly created

yannickroger:
Link: https://jira.ez.no/browse/EZP-24100

## Description
When using `urlalias_iri` images are renamed with the object name. The limit for object names is by default 255 UTF-8 char (and can be customized). The problem is that most file systems have a 255 bytes limit so they won't be able to create the image file.

This patch introduces a 200 bytes limit for images (leaving a bit of room for alias name and extension).

## Tests
Manual and unit tests

## TODO
* [x] Maybe change the public method to private, but need to find a way to unit test it.
* [x] Rebase/squash commits once accepted

### DIFF
--- a/kernel/classes/datatypes/ezimage/ezimagealiashandler.php
+++ b/kernel/classes/datatypes/ezimage/ezimagealiashandler.php
@@ -269,7 +269,7 @@ class eZImageAliasHandler
             }
         }
         $objectName = eZImageAliasHandler::normalizeImageName( $objectName );
-        $objectName .= $this->imageSerialNumber();
+        $objectName = $this->trimToFileSystemFileName( $objectName ) . $this->imageSerialNumber();
 
         return $objectName;
     }
@@ -304,7 +304,7 @@ class eZImageAliasHandler
             }
         }
         $objectName = eZImageAliasHandler::normalizeImageName( $objectName );
-        return $objectName;
+        return $this->trimToFileSystemFileName( $objectName );
     }
 
     /*!
@@ -378,11 +378,34 @@ class eZImageAliasHandler
         $pathParts = array( eZSys::storageDirectory(), $contentImageSubtree );
         if ( $pathString != '' )
         {
-            $pathParts[] = $pathString;
+            //Make sure that all folders are smaller than the FS limit
+            foreach ( explode( '/', $pathString ) as $folder )
+            {
+                $pathParts[] = $this->trimToFileSystemFileName( $folder );
+            }
         }
         $pathParts[] = $attributeID . '-' . $attributeVersion . '-' . $attributeLanguage;
         $imagePath = implode( '/', $pathParts );
         return $imagePath;
+    }
+
+    /**
+     * Trims a filename to a given limit (in bytes).
+     * It can be used to avoid file system limitations
+     *
+     * @param string $longName
+     * @param int $limit in bytes
+     *
+     * @return string shortened name
+     */
+    private function trimToFileSystemFileName( $longName, $limit = 200 )
+    {
+        if ( strlen( $longName ) <= $limit )
+        {
+            return $longName;
+        }
+
+        return mb_strcut( $longName, 0, $limit, "utf-8" );
     }
 
     /*!
--- a/tests/tests/kernel/datatypes/ezimage/ezimagealiashandler_regression.php
+++ b/tests/tests/kernel/datatypes/ezimage/ezimagealiashandler_regression.php
@@ -12,12 +12,6 @@ class eZImageAliasHandlerRegression extends ezpDatabaseTestCase
 {
     protected $backupGlobals = false;
 
-    public function __construct()
-    {
-        parent::__construct();
-        $this->setName( "eZImageAliasHandler Regression Tests" );
-    }
-
     /**
      * Regression test for issue #15155
      *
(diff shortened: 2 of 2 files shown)


tests/tests/kernel/datatypes/ezimage/ezimagealiashandler_regression.php
```diff
@@ -12,12 +12,6 @@ class eZImageAliasHandlerRegression extends ezpDatabaseTestCase
 {
     protected $backupGlobals = false;
 
-    public function __construct()
-    {
-        parent::__construct();
-        $this->setName( "eZImageAliasHandler Regression Tests" );
-    }
-
```
yannickroger: The constructor what breaking the `@dataProvider`. Is setting the name really usefull.

kernel/classes/datatypes/ezimage/ezimagealiashandler.php
```diff
@@ -378,13 +378,36 @@ function imagePathByNode( $contentObjectAttribute, $mainNode )
         $pathParts = array( eZSys::storageDirectory(), $contentImageSubtree );
         if ( $pathString != '' )
         {
-            $pathParts[] = $pathString;
+            //Make sur that all folders are smaller than the FS limit
```
crevillo: sure :)

yannickroger:
@crevillo done :) Thanks

kernel/classes/datatypes/ezimage/ezimagealiashandler.php
```diff
@@ -378,13 +378,36 @@ function imagePathByNode( $contentObjectAttribute, $mainNode )
...
+    /**
+     * Trims a filename to a given limit (in bytes).
+     * It can be used to avoid file system limitations
+     *
+     * @param $longName
+     * @param int $limit in bytes
+     *
+     * @return string shortened name
+     */
+    public static function trimToFileSystemFileName( $longName, $limit = 200 )
```
bdunogier: Unless I've missed something, you don't have to make it static (but it should be public). `imageNameByNode` already uses `$this`, and `imagePathByNode` isn't called statically:

```
$ git grep imagePathByNode
kernel/classes/datatypes/ezimage/ezimagealiashandler.php:    function imagePathByNode( $contentObjectAttribute, $mainNode )
kernel/classes/datatypes/ezimage/ezimagetype.php:                $dirpath = $imageHandler->imagePathByNode( $contentObjectAttribute, $mainNode );
```

tests/tests/kernel/datatypes/ezimage/ezimagealiashandler_regression.php
bdunogier: You just need to pass along the parent constructor arguments, Yannick.

No, the name isn't useful, but it's easy to fix :-)

tests/tests/kernel/datatypes/ezimage/ezimagealiashandler_regression.php
```diff
@@ -88,6 +82,38 @@ public function testIssue15155()
...
+            )
+        );
+    }
+
+    /**
+     * @dataProvider providerForTestTrimToFileSystemFileName
+     */
+    public function testTrimToFileSystemFileName( $longName, $limit, $expects )
+    {
+        $this->assertEquals( $expects, eZImageAliasHandler::trimToFileSystemFileName( $longName, $limit ) );
```
bdunogier: nitpick: e test would make more sense if it was length based, and not string based. We don't really care that the returned string is equal to "私aはb簡c単dにe", we care that it is not longer than `$limit`.

yannickroger: Actually since we are spliting UTF-8 chars, triming to 20 bytes a string can provide a 18 bytes lenght string.

bdunogier: 18 is < 20, it would pass :)

kernel/classes/datatypes/ezimage/ezimagealiashandler.php
```diff
@@ -378,13 +378,36 @@ function imagePathByNode( $contentObjectAttribute, $mainNode )
...
         $pathParts[] = $attributeID . '-' . $attributeVersion . '-' . $attributeLanguage;
         $imagePath = implode( '/', $pathParts );
         return $imagePath;
     }
 
+    /**
+     * Trims a filename to a given limit (in bytes).
+     * It can be used to avoid file system limitations
+     *
+     * @param $longName
```
lolautruche: Missing type

kernel/classes/datatypes/ezimage/ezimagealiashandler.php
```diff
@@ -378,13 +378,36 @@ function imagePathByNode( $contentObjectAttribute, $mainNode )
...
+     * @param int $limit in bytes
+     *
+     * @return string shortened name
+     */
+    public static function trimToFileSystemFileName( $longName, $limit = 200 )
+    {
+        if ( strlen( $longName ) <= $limit )
```
lolautruche: Better use `mb_strlen()`

yannickroger: `mb_strlen()` is causing the issue :) Bytes (File system) vs Char

bdunogier: What Yannick means is that the issue is the space we have in the database, and this is counted as non-multibyte. We want to truncate the string that (non-mb) length, not to its true length.

tests/tests/kernel/datatypes/ezimage/ezimagealiashandler_regression.php
glye: Well, what if a later regression makes it return an empty string? That would pass too. I like that it checks for sensible content. (Could do both.)

glye:
Some doubts about the tests, but +1

yannickroger:
Method changed to `private` and unit test updated to test parent methods.
Open for review.

andrerom:
+1, but is new stack affected by this as well? (ping @lolautruche )

lolautruche:
Don't know, if it's the case then it's IOService job. Ping @bdunogier :-)

bdunogier:
Might very well be. I don't see any reason why it would not.

Good occasion for @yannickroger to get into the new stack :-)

lolautruche:
Touché :stuck_out_tongue_winking_eye:

bdunogier:
> Touché

(touché)[http://media.giphy.com/media/i5J5N8ZAlStlm/giphy.gif] ?

bdunogier:
Touché like this ?

!(http://media.giphy.com/media/i5J5N8ZAlStlm/giphy.gif)

yannickroger:
@lolautruche @bdunogier and about reviewing this PR ?
Regarding fixing it in the new stack, I don't mind, as soon as I am done helping out on the platform UI.

andrerom:
@yannickroger you have 2x +1 now ;)

yannickroger:
Travis fails while it worked on my local env, any idea what could cause this ?

tests/tests/kernel/datatypes/ezimage/ezimagealiashandler_regression.php
```diff
@@ -88,6 +82,83 @@ public function testIssue15155()
...
+     * @dataProvider providerForTestImageNameObjectFalse
+     */
+    public function testImageNameObjectFalse( $longName, $expects )
+    {
+        ezpINIHelper::setINISetting( 'site.ini', 'URLTranslator', 'TransformationGroup', 'urlalias_iri' );
+
+        $handler = new eZImageAliasHandler( null );
+        $language = "fre-FR";
+        $contentVersionMock = $this->getMockBuilder( 'eZContentObjectVersion' )->disableOriginalConstructor()->getMock();
+        $contentVersionMock->expects( $this->any() )->method( 'versionName' )->willReturn( $longName );
```
lolautruche: use `will()` instead of `willReturn()` (not compatible with required PHPUnit version in legacy)

PR #1153, opened 2015-03-27
Fix EZP-24197: DOC: ezcache.php --expiry parameter can only be used together with --purge

glye:
Add doc about this to --help
Refuse to accept --expiry with --purge

https://jira.ez.no/browse/EZP-24197


yannickroger:
+1

bdunogier:
+1

PR #1154, opened 2015-03-27
Fix EZP-24141: Double slash in url after having searched a default locat...

glye:
...ion for an object relation in class edit module

If there are any unordered parameters, remove trailing slash(es) before adding the parameters, to avoid double slash.

The secondary complaint, that `language` occurs twice in URLs like `/class/edit/45/language/fre-FR//(language)/fre-FR`, is not fixed. It doesn't cause any problems for the issue reporter, and would require quite some rewriting of the class edit view, so I don't think it's worthwhile given that it's legacy.

https://jira.ez.no/browse/EZP-24141

### DIFF
--- a/lib/ezutils/classes/ezmodule.php
+++ b/lib/ezutils/classes/ezmodule.php
@@ -857,14 +857,17 @@ class eZModule
             $uri .= $unorderedURI;
         }
 
-        if ( is_array( $userParameters ) )
+        if ( !empty( $userParameters ) && is_array( $userParameters ) )
         {
+            // Remove any slash(es) at the end of the uri, because the user params are appended with a leading slash
+            $uri = preg_replace( "#(^.*)(/+)$#", "\$1", $uri );
             foreach ( $userParameters as $name => $value )
             {
                 $uri .= '/(' . $name . ')/' . $value;
             }
         }
 
+        // Remove multiple slashes at the end of the uri
         $uri = preg_replace( "#(^.*)(//+)$#", "\$1", $uri );
         if ( $anchor !== false )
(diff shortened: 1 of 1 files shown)


lib/ezutils/classes/ezmodule.php
```diff
@@ -857,8 +857,9 @@ function redirectionURIForModule( $module, $viewName, $parameters = array(),
             $uri .= $unorderedURI;
         }
 
-        if ( is_array( $userParameters ) )
+        if ( is_array( $userParameters ) && count( $userParameters ) > 0 )
```
andrerom: nitpick: you should also be able to use `!empty()`, and put it before `is_array`. it's language construct, so vastly faster.

andrerom:
As for language doucle in url, I agree that should be reported as separate improvement and closed as won't fix.

glye:
Thanks, using empty() now.

andrerom:
I'm +1

if you can add some inline php doc on what is going on here.
For instance I'm not sure I follow why the one after is different, ref `$uri = preg_replace( "#(^.*)(//+)$#", "\$1", $uri );`

glye:
Inline doc added. I'm not sure either why the one below is different, though :-)
It was done in https://jira.ez.no/browse/EZP-7054 / https://github.com/glye/ezpublish-legacy/commit/205146612a5c27661281c879799191cf2205d506
where it was changed from "remove any slashes at the end of uri" to "remove multiple slashes at the end of uri; if there is only one slash, don't touch it".

yannickroger:
+1
(don't forget to squash commits)

PR #1156, opened 2015-04-08
Fix EZP-24214: Incorrect Finnish country name translation: Suomi (ezcountry datatype)

glye:
Not much to say here.

https://jira.ez.no/browse/EZP-24214


crevillo:
+1, even i don't know Finnish :).

janit:
+1 That is the official name seldomly used.

bdunogier:
My Finnish is a bit rusty, but it looks legit.

+1

PR #1157, opened 2015-04-08
Fix EZP-24233: copysubtree: wrap object copy in transaction.

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-24223

When using MySQL master/slave setup, copysubtree will fail with a fatal error (`Call to a member function attribute() on a non-object  ...`)
This happens because a new object version is created and immediately read from (on the slave server).

Fixed by wrapping relevant section in a DB transaction.

### DIFF
--- a/kernel/content/copysubtree.php
+++ b/kernel/content/copysubtree.php
@@ -167,6 +167,9 @@ function copyPublishContentObject( $sourceObject,
         return 0;
     }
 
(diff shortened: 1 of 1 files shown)


andrerom:
+1

yannickroger:
+1
Thanks for the good description. Ready for merge if you have time @joaoinacio

PR #1160, opened 2015-04-17
Fix EZP-24233: undisclosed-recipients:; breaks sendmail

yannickroger:
Link: https://jira.ez.no/browse/EZP-24233
You can also have a look at the (old): https://jira.ez.no/browse/EZP-13617

## Description
Sendmail (the MTA) doesn't support `undisclosed-recipients:;` in the `TO:` of the header. It doesn't seem to be a standard to have it there according to https://tools.ietf.org/html/rfc5322#section-3.6.3

This patch introduces a new setting. I know we are trying to avoid it, but the goal is to avoid BC breaks as much as possible. This setting allows to replace the `undisclosed-recipients:;` by an empty string.

## Test
I am not using sendmail as MTA. So I made sure  `undisclosed-recipients:;`  is removed (or not) from the headers and I did sanity tests using my MTA (exim4). It has been tested by somebody else using sendmail.

## TODO
* [x] : Have somebody test it on sendmail

### DIFF
--- a/lib/ezutils/classes/ezsendmailtransport.php
+++ b/lib/ezutils/classes/ezsendmailtransport.php
@@ -65,7 +65,9 @@ class eZSendmailTransport extends eZMailTransport
             if ( $sys->osType() != 'win32' )
             {
                 $excludeHeaders[] = 'To';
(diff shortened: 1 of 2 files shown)


lolautruche:
+1

andrerom:
+1

PR #1164, opened 2015-04-28
Fix EZP-19256: eZ JS packer should insert semicolons as needed

sterrien:
https://jira.ez.no/browse/EZP-19256

Add a semicolon after each JS file.

### DIFF
--- a/extension/ezjscore/classes/ezjscpacker.php
+++ b/extension/ezjscore/classes/ezjscpacker.php
@@ -429,6 +429,7 @@ class ezjscPacker
         // Merge file content and create new cache file
         $content = '';
         $isCSS = $data['file_extension'] === '.css';
+        $isJS  = $data['file_extension'] === '.js';
         foreach( $data['locale'] as $i => $file )
         {
             // if this is a js / css generator, call to get content
@@ -461,6 +462,10 @@ class ezjscPacker
 
             $content .= "/* start: $file */\r\n";
             $content .= $fileContent;
+            if ( $isJS )
+            {
+                $content .= ';';
+            }
             $content .= "\r\n/* end: $file */\r\n\r\n";
         }
 


brookinsconsulting:
Hello @sterrien 

This looks mighty clean (from a cs stand point) :)

The commit message convention would imply that you rename your commit message to be the following instead, "Fix EZP-19256: eZ JS packer should insert semicolons as needed" (without the quotes).

Cheers,
Brookins Consulting

sterrien:
done !

brookinsconsulting:
Hello @sterrien 

You rock! Thank you again for your continued support!

Cheers,
Brookins Consulting

sterrien:
@brookinsconsulting 
It's a pleasure. I'm just waiting for someone who will accept the PR ;-)
Do you know, approximately, how long will it take ?

brookinsconsulting:
Hello @sterrien 

Well in general PRs take two +1(s) from senior developers on the project (from what I've heard).

The time it takes to review, test, discuss and approve a PR is completely variable and subject to many concerns. In short there is not fixed length of time.

In this case (with legacy PRs) there is less focus on these since the introduction of ezplatform.
I am trying to step in to help continue the PR process for legacy PRs but time is always a concern.

I spent time tonight (instead of sleeping; again) to review your changes in greater detail, the issue ticket for the PR and test the changes.

I hesitate to say more or give a plus one on this one for reasons that **will** get me into trouble no matter what I say. On one side the syntax elitists will insist that any valid js be treated properly and on the other side of the best practices people will insist that if you can not be trusted to use the standards used almost universally in so many (not all, but many) languages like consistent line endings (ie: use of ';' char to indicate line endings) then you have further and greater problem long term than to even bother continuing a conversation. 

This this issue (which is not yours mind you) is pure flame bait and in the past I would avoid speaking up and keep my head down to avoid the cross fire. Also in the past I have seen the suggested fixes (like the one in the issue) simply ignored by eZ Systems (which it was, till now) as a 'hack' which does not truly or properly solve the problem ... the right way or best way possible.

I want to encourage your efforts but at the same time, I personally have both conflicted feelings about the very nature of this PR and even more with the changes suggested.

In short I am of the opinion, after even only the most basic research and testing, that the suggested changes in the issue ticket ... are in-deed a hack which while in some use cases might solve the issue presented, in general, how the problem is solved (again in the ticket) is the wrong way to implement a solution to a much larger problem and more specifically the changes fail in my very first blind test code. 

I was not trying to break the fix but without copy and paste exact test case javascript to test ... I did have to write my own which the presented fix failed to solve the problem. 

In short this is a hack and is a dirty one at best. The real solution is to implement something much much more complicated which does the real work required to detect missing ';' chars and add them on the fly (with exacting / proper placement regardless of artifacts) in PHP which would be much much more work, cost much more in terms of run time performance and ultimately address code from poor programers who insist on writing code which is in my opinion non-standard. When I say non-standard I mean, it's legal, standalone it could run but that does not make it standard or right in the eyes of the majority of programers, existing code and strongly recommended best practices who do follow the conventions which help retain the sanity of everyone who comes into contact with their code. Which I personally feel is a 'code standards' concern.

I would not willingly accept code from another which did not meet the code standards requirements of my project .. for js, for me (today) line endings (ie: js without ';' chars) violates those code standards and I also would not support the code since I would not accept it or use it. 

So .. adding support to a deprecated cms framework like legacy to add a barely tested hack which in some test cases works in adding missing ';' chars to line endings of the end of file contents, while adding the char without testing for the need first (my frustration at first glance when you posted this PR) seems to me like supporting a violation of my best practices and code standards all while ensuring that the users of ezjscore in the future would have ugly 'empty statements' with ';;' (which I loath even at the cost of being labeled some derogatory term by a syntax purest who insists there is no problem because ';;' is valid javascript) all because someone on internet wants to insist on breaking best practices in the name of 'valid js should be good enough for everyone in every use case' and use non-standard javascript without line endings with ezjscore which it's use is not required (just enabled by default; hardly a concern for the distribution package as kernel overrides provide for overriding this class by default, cleanly without kernel hacking). It all just rubs me the wrong way, and now, "I said too much, but I haven't said enough" (REM:Losing My Religion) ... *intro the hate mail*

Because of the incomplete and non specific 'steps to reproduce' I did what was asked in those instructions (basically) and still found that the code I tested, would not be solved by the code changes in question. Here is the code for custom1.js and custom2.js that I tested first before starting this reply.

*extension/customjs/design/admin2/javascript/custom1.js*
```javascript
// define a function                                                                                                                                                                               
var fn = function () {
    console.log( 'function has run' ) // semicolon missing at this line
} // semicolon missing at this line
```
*extension/customjs/design/admin2/javascript/custom2.js*
```javascript
// then execute some code inside a closure                                                                                                                                                         
(function () {
    //...                                                                                                                                                                                          
})();
```

Also worth noting the steps to reproduce are using deprecated and version specific ini settings so for the record here is the ini settings I used to include the two custom js files.

*settings/siteaccess/ezdemo_site_admin/design.ini.append.php*
```ini
[JavaScriptSettings]
BackendJavaScriptList[]=custom1.js
BackendJavaScriptList[]=custom2.js
```

The problem I quickly realized with my test code was that the issue ticket fix, now recognized as a static hack which only works in a very very specific test / use case and nothing more. As I mentioned before this change would never be accepted in the past in it's current state and the issue would have to be expanded in scope and detail before a much more responsible and feature complete solution to the problem real issue would be implemented instead. 

In this case I'm not sure of a quick way to replace the hack with a much more complete solution to the problem. In short I think to solve this problem we would need to implement js missing semicolon detection; properly and then semicolon insertion (regardless of comments or anything code uniqueness conflicts like I previously found I could produce on my first try without trying). I hear that jslint is very good at detecting missing ';' characters but then that's not a PHP based GPLv2 compatible licensed solution we even could make use of outright even though, by legacy convention, if we implement a feature, we normally write our own implementation of the code required, often from scratch.

But overall the simpler answer here that may not be liked by all, most hate this answer .. the real answer is to jslint test your javascript code and write javascript code which conforms to best practices, coding conventions and for the love of g*d, sanity before you try to use it within a real website based on ezjscore. Worth noting I could not find a jslint option to even ignore missing semicolon errors! Now that spoke to me and I think it should speak volumes to others who say semicolons are not required ... I also could not find a non js based js validator but then I kinda got sleepy after so many hours writing this reply after so much review, testing and careful consideration. 

I did force myself to do the work required to properly give a first round review of your PR (in my eyes) regardless of what it costs me (sleep, sway, friends who don't write js with semicolons). I spent the time, I did not ignore this PR but that said .. 

In it's current form I see no option here, I must, on principle vote, -1 on this PR (in it's current form). And for this I am sorry, for you since I want badly to support your desire to help but not sorry I do not wish to support hacks like this within ezjscore / ezpublish-legacy.

I wish the issue ticket in question would be closed as "Won't fix" / "Not a bug" / "Don't use code that does not pass jslint 100% error free".

Cheers,
Brookins Consulting

andrerom:
*( side:  @brookinsconsulting you should write everything over 3 paragraphs as blog posts maybe and refer to it and reuse it, might come in handy and save you and readers some time )*

+1, one more review and it can be merged

PR #1165, opened 2015-05-05
Fix EZP-24333: Fatal error importing ezpackage generated with ezpm.php

joaoinacio:
https://jira.ez.no/browse/EZP-24333

#### Problem: ####
Packages generated by ezpm.php may not include some xml elements (such as type or description),
though ezpackage import seems to require/expect these in the xml structure.

#### Fix: ####
Do not assume the elements exist when parsing the xml structure, as they are not *absolutely* required.

#### Tests: ####
Manual.


andrerom:
+1

brookinsconsulting:
+1

glye:
+1

PR #1166, opened 2015-05-08
Fix EZP-24240: links applied incorrectly when text has line break.

joaoinacio:

#### Problem: ####
When applying links in ezoe to a multi-line text (using `<shift+enter>`), multiple `link` tags are created.
The first link is correct, however the second/other links have an incorrect id attribute "`__mce_tmp`" which is left unreplaced.

After this, adding a different link to another element in ezoe will replace the anchor of the existing link with id "`__mce_tmp`".

#### Solution: ####
When inserting links into multiple lines/elements, apply attributes to all the "__mce_tmp" instances.

### DIFF
--- a/extension/ezoe/design/standard/javascript/ezoe/popup_utils.js
+++ b/extension/ezoe/design/standard/javascript/ezoe/popup_utils.js
@@ -255,8 +255,14 @@ var eZOEPopupUtils = {
                 if ( n && n.nodeName )
                     s.editorElement = n;
             }
-            else
+            else {
                 ed.dom.setAttribs( s.editorElement, args );
+                // apply attributes to all '__mce_tmp' elements
+                if ( s.tagName == 'link' ) do {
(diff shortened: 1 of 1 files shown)


joaoinacio:
Note:
Changes from the ezoe popup seem to apply to a single element, by design, which isn't the case with multi-lines.
If necessary, suggestions on how to further cleanup/refactor relevant code are welcome :)

andrerom:
I guess the patch is ok, however feels like a workaround.
Most likely we should not allow adding links across several lines, however that would cause some issues as well. /cc @dpobel

joaoinacio:
The problem is that `editorElement` must not be the whole `p`aragraph (or div,span,etc) but each selected text line - IMHO removing this feature would be overkill...
@andrerom, @dpobel : Indeed it doesn't look/feel like a perfect solution, in this case it may just be, afaict, the "lesser evil" - alternatives/suggestions are welcome.

yannickroger:
closing in favor of #1176

PR #1170, opened 2015-05-14
Fix trimming DOMElement textContent properties (re-enables PHP 5.6 on Travis)

jeromegamez:
The PHPUnit tests currently fail when running on PHP 5.6, which seems to… different on `DOMNode`'s read-only properties than previous reasons.

Currently, e.g. the `eZXMLTextRegression::testNonBreakingSpace()` would fail, because `esp&nbsp;ace` gets "trimmed" to `esp&amp;nbsp;ace`, which is happening in [eZSimplifiedXMLInputParser::structHandlerText()](https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/classes/datatypes/ezxmltext/handlers/input/ezsimplifiedxmlinputparser.php#L687)

```php
$element->textContent = ltrim( $element->textContent );
```

`$element->textContent` is a read-only property, as [documented here](http://php.net/manual/en/class.domnode.php).

I don't know why PHP does some HTML entity encoding (that's weird) instead of emitting an error or a warning, but whatever :).

This PR fixes the issue by replacing the overwriting of the property with a recreation of the TextNode.

The tests now run through on my local PHP 5.6 environment, and I hope the Travis-CI tests to as well for the other PHP versions.

Cheers
:octocat: Jérôme

### DIFF
--- a/extension/ezoe/ezxmltext/handlers/input/ezoeinputparser.php
+++ b/extension/ezoe/ezxmltext/handlers/input/ezoeinputparser.php
@@ -951,11 +951,16 @@ class eZOEInputParser extends eZXMLInputParser
             if ( $trim )
             {
                 // Trim and remove if empty
-                $element->textContent = ltrim( $element->textContent );
-                if ( $element->textContent == '' )
+                $parent = $element->parentNode;
+                $trimmedElement = new DOMText( ltrim( $element->textContent ) );
+
+                if ( $trimmedElement->textContent == '' )
                 {
-                    $parent = $element->parentNode;
-                    $element = $parent->removeChild( $element );
+                    $parent->removeChild( $element );
+                }
+                else
+                {
+                    $parent->replaceChild( $trimmedElement, $element );
                 }
             }
         }
--- a/kernel/classes/datatypes/ezxmltext/handlers/input/ezsimplifiedxmlinputparser.php
+++ b/kernel/classes/datatypes/ezxmltext/handlers/input/ezsimplifiedxmlinputparser.php
@@ -684,11 +684,16 @@ class eZSimplifiedXMLInputParser extends eZXMLInputParser
             if ( $trim )
             {
                 // Trim and remove if empty
-                $element->textContent = ltrim( $element->textContent );
-                if ( $element->textContent == '' )
+                $parent = $element->parentNode;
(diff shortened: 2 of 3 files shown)


jeromegamez:
Yeah!

kernel/classes/datatypes/ezxmltext/handlers/input/ezsimplifiedxmlinputparser.php
```diff
@@ -684,11 +684,16 @@ function structHandlerText( $element, &$newParent )
...
+
+                if ( $trimmedElement->textContent == '' )
                 {
-                    $parent = $element->parentNode;
-                    $element = $parent->removeChild( $element );
+                    $parent->removeChild( $element );
+                }
+                else
+                {
+                    $parent->replaceChild($trimmedElement, $element);
```
crevillo: Cs as above.

extension/ezoe/ezxmltext/handlers/input/ezoeinputparser.php
```diff
@@ -951,11 +951,16 @@ function structHandlerText( $element, $newParent )
...
+
+                if ( $trimmedElement->textContent == '' )
                 {
-                    $parent = $element->parentNode;
-                    $element = $parent->removeChild( $element );
+                    $parent->removeChild( $element );
+                }
+                else
+                {
+                    $parent->replaceChild($trimmedElement, $element);
```
crevillo: Cs :)

jeromegamez: Fixed

kernel/classes/datatypes/ezxmltext/handlers/input/ezsimplifiedxmlinputparser.php
jeromegamez: Fixed

lolautruche:
+1

Review ping @bdunogier @dpobel @yannickroger @andrerom

andrerom:
+1 :tada:

lolautruche:
Thanks @jeromegamez !

Should that be backported @andrerom ?

andrerom:
yes, at least to 5.4, maybe as PR so we can see if shippable is able to comply.

PR #1171, opened 2015-05-20
Fix EZP-16874: view cache file written when cache_ttl=0 is set in tem…

fwoldt:
Fix the issue: 
if  {set-block scope=global variable=cache_ttl}0{/set-block} is set in view full template, 
no cache file should be generated to reduce io write calls and storage size

### DIFF
--- a/kernel/classes/eznodeviewfunctions.php
+++ b/kernel/classes/eznodeviewfunctions.php
@@ -605,6 +605,12 @@ class eZNodeviewfunctions
             $validation
         );
 
+        // by setting cache_ttl to 0 in tpl code no cachfile should be generated
+        if ( isset( $result['no_cache'] ) and $result['no_cache'] )
+        {
+            $noCache = true;
+        }
+
         // 'store' depends on noCache: if $noCache is set, this means that retrieve
         // returned it, and the noCache fake cache file is already stored
         // and should not be stored again


kernel/classes/eznodeviewfunctions.php
```diff
@@ -605,6 +605,12 @@ static public function contentViewGenerate( $file, $args )
             $validation
         );
 
+        // by setting cache_ttl to 0 in tpl code no cachfile should be generated
+        if ( isset( $result['no_cache'] ) and $result['no_cache'] )
```
crevillo: symbols based boolean operators must be used according to coding standards.  can you change the```and``` to ```&&```?

fwoldt: thank's for the tip - i fixed it 1 minute ago ;-)

kernel/classes/eznodeviewfunctions.php
```diff
@@ -605,6 +605,12 @@ static public function contentViewGenerate( $file, $args )
             $validation
         );
 
+        // by setting cache_ttl to 0 in tpl code no cachfile should be generated
```
gggeek: nitpick: cache file

andrerom:
I'm not sure I remember the details right, but I think we changed it to write a file to avoid cache_ttl=0 taking down the servers. Cause if the file is there advising no cache, then no lock will be created while generating cache. But it is a long time ago. @bdunogier might remember, but git history might also.

fwoldt:
@andrerom if the viewcache file is needed for performance reasons, why not use one empty temp cachefile for all viewcaches wiht cache_ttl=0?

andrerom:
It can probably be optimized to not store anything else then the noCache flag and other relevant meta data indeed. Changes to all handlers and cluster handler code is probably needed for that.

efernandes:
Afaik, the cache generation was intentional and comes from https://jira.ez.no/browse/EZP-14850

PR #1172, opened 2015-05-21
Fixing HTTP ContentLanguage

yannoz:
Some HTTP ContentLanguage values seem to be quite inaccurate.

I tried to fix them using the following ressources:
http://www.metamodpro.com/browser-language-codes
http://www.iana.org/assignments/language-subtag-registry/language-subtag-registry

By the way, the `ser-SR.ini` should be renammed `ser-RS` (since SR is actually the country code of Suriname), but it might break existing installations using this locale if the database is not updated accordingly.


trolek:
There is another PR to this issue: #676

PR #1173, opened 2015-05-28
Fix EZP-24419: clear stale active_extension cache files.

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-24419

##### Problem: #####
Cache files for active_extensions aren't removed when clearing/purging caches.

##### Solution: #####
Make sure to remove them :)

Note that `ClusterFileHandler` isn't used here, also active extension cache is hardcoded to var/cache.
As no record of files is kept, the proposed solution is to have eZExtension clean up the stale files much in the same way as it creates them: manually (see [ezextension.php#L115](https://github.com/ezsystems/ezpublish-legacy/blob/master/lib/ezutils/classes/ezextension.php#L115) ).

##### Tests: #####
Manual.

### DIFF
--- a/kernel/classes/ezcache.php
+++ b/kernel/classes/ezcache.php
@@ -781,6 +781,7 @@ class eZCache
         $handler->setTimestamp( $cacheItem['expiry-key'], time() );
         $handler->store();
 
+        eZExtension::clearActiveExtensionsCache();
         eZExtension::clearActiveExtensionsMemoryCache();
     }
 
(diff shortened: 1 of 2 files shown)


andrerom:
Another option is to expire it, as the generation code does check expiry timestamp, any reason why you went this route?

joaoinacio:
@andrerom: the problem is that expiry is already being done (through the timestamp in expiry.php ), but there is nothing actually removing/purging the stale files, so they will eventually start to pile up.

A possible improvement/alternative would be to use a separate purge function for this, although it doesn't seem necessary AFAICT (much like ini cache, etc)

andrerom:
ok, well +1, it's ok like this.

lolautruche:
+1

PR #1176, opened 2015-06-08
Fix EZP-24240: Incorrect links when applied on multiple lines

dpobel:

# Description

It's an alternative to https://github.com/ezsystems/ezpublish-legacy/pull/1166. Instead of fixing the potential multiple remaining ids afterwards, this patch uses a special class and make sure to clean up right after we don't need it anymore.

# Tests

manual tests


dpobel:
ping @yannickroger @andrerom @joaoinacio

yannickroger:
+1

andrerom:
+1

PR #1177, opened 2015-06-18
Fix EZP-24334: Error exporting ezpackage with ezpm.php through ezpublish:legacy:script

lolautruche:
> https://jira.ez.no/browse/EZP-24334

Some scripts, like `ezpm.php`, initialize `eZScript` with session support.
In that case, when shutting down, `eZScript` tries to log out current user and to clean up the session. This calls `ezpSessionHandlerSymfony`, which calls Symfony session storage, which does... a legacy callback.

This patch ensures that calls to session regenerate/destroy/gc in `ezpSessionHandlerSymfony`, and  made from CLI doesn't do anything.

AFAICS, there should not be any BC break.

### DIFF
--- a/lib/ezsession/classes/ezpsessionhandlersymfony.php
+++ b/lib/ezsession/classes/ezpsessionhandlersymfony.php
@@ -45,6 +45,11 @@ class ezpSessionHandlerSymfony extends ezpSessionHandler
 
     public function destroy( $sessionId )
     {
+        if ( eZSys::isShellExecution() )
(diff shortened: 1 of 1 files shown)


lolautruche:
Review ping @andrerom @bdunogier @yannickroger @pspanja @glye

yannickroger:
+1

glye:
+1 if it works :)

pspanja:
+1

PR #1178, opened 2015-06-26
EZP-24240: Make sure the custom temporary class is not added back on links

dpobel:

# Description

The fix for [EZP-24240](https://jira.ez.no/browse/EZP-24240) provided in https://github.com/ezsystems/ezpublish-legacy/pull/1176 was incomplete. This patch makes sure we don't re-add the custom temporary class `ezoeInsertedLink` otherwise if you create several links in a row, they'll all have the target of the last one.

# Tests

manual tests


andrerom:
+1

yannickroger:
+1

PR #1179, opened 2015-06-29
Fix EZP-24492: link attributes unusable with custom tag containing underscores

dpobel:
JIRA: https://jira.ez.no/browse/EZP-24492

# Description

The "link" attribute of a custom tag can be unusable if the custom tag name contains an underscore. This is happening because several elements ids are computed by splitting a string containing the tag name with the underscore as a separator.

This patch makes sure we don't take the tag name into account when computing the elements ids.

# Tests

manual tests (on the link tag and custom tags with and without underscores)


yannickroger:
+1
ping @andrerom

andrerom:
+1

PR #1180, opened 2015-06-29
Prevent rss export crash when content's creator object does not exist…

alafon:
… anymore

### DIFF
--- a/kernel/classes/ezrssexport.php
+++ b/kernel/classes/ezrssexport.php
@@ -449,6 +449,10 @@ class eZRSSExport extends eZPersistentObject
                 $creatorObject->attribute('name'), ENT_NOQUOTES, 'UTF-8'
             );
         }
+        else
+        {
+            $author->name = "Admin from " . $config->variable( 'SiteSettings', 'SiteName' );
+        }
 
         $imageURL = $this->fetchImageURL();
         if ( $imageURL !== false )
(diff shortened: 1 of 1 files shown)


kernel/classes/ezrssexport.php
```diff
@@ -449,6 +449,10 @@ function generateFeed( $type )
                 $creatorObject->attribute('name'), ENT_NOQUOTES, 'UTF-8'
             );
         }
+        else
+        {
+            $author->name = "Admin from " . $config->variable( 'SiteSettings', 'SiteName' );
```
yannickroger: We need a dedicated translation string.
"Admin of" should be better ?

kernel/classes/ezrssexport.php
```diff
@@ -567,6 +571,12 @@ function generateFeed( $type )
                     );
                     $author->email = $config->variable( 'MailSettings', 'AdminEmail' );
                 }
+                else
+                {
+                    $author = $item->add('author');
+                    $author->name = "Admin from " . $config->variable( 'SiteSettings', 'SiteName' );
```
yannickroger: same as above.

yannickroger:
+1 with remarks taken into account and an jira issue (also update the commit message with the issue number).

kernel/classes/ezrssexport.php
alafon: Actually this was more exemple than anything... 

I'm not even sure if using "Admin" is good.
I've thought of using :
- the Lastname/Firstname of the admin user
- the AdminMail from the site.ini 

Using the AdminMail would duplicate `$author->email `, but, if you consider that this is more a workaround than a bugfix... well, I think that we won't find something which will fit for all the cases (AdminMail from .ini is often fake as well as the AdminUser which is not always related to a real person...).

What do you think ?

yannickroger: Avoiding from crashing is great :)
I have no preference regarding the fallback message as long as we can customize it a bit. We are dealing with an edge case as user should not be deleted in the first place.

PR #1181, opened 2015-06-29
Fix EZP-24521: Performance in ezxmltexttype attribute deletion

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-24521

In eZContentObject::purge(),  for every attribute attribute (for every version) the deleteStoredObjectAttribute() function is called.

For contents with multiple versions, removing the urls linked to ezxml contents can have an impact on performance.

This optimizes `eZXMLTextType::deleteStoredObjectAttribute()` by executing this operation only once per attribute when no version is given (removing data for all versions).

### DIFF
--- a/kernel/classes/datatypes/ezxmltext/ezxmltexttype.php
+++ b/kernel/classes/datatypes/ezxmltext/ezxmltexttype.php
@@ -99,6 +99,7 @@ class eZXMLTextType extends eZDataType
     {
         $this->eZDataType( self::DATA_TYPE_STRING, ezpI18n::tr( 'kernel/classes/datatypes', "XML block", 'Datatype name' ),
                            array( 'serialize_supported' => true ) );
+        $this->deletedStoredObjectAttribute = array();
     }
 
     /*!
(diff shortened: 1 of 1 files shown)


kernel/classes/datatypes/ezxmltext/ezxmltexttype.php
```diff
@@ -818,6 +822,12 @@ function deleteStoredObjectAttribute( $contentObjectAttribute, $version = null )
 
             $db->query( "DELETE FROM ezurl WHERE id IN ($unusedUrlIDString)" );
         }
+
+        /* If all the versions/urls of the attribute where removed, do not try to remove them again */
```
joaoinacio: Nitpick: typo (were removed)

kernel/classes/datatypes/ezxmltext/ezxmltexttype.php
```diff
@@ -843,6 +853,8 @@ function batchInitializeObjectAttributeData( $classAttribute )
         $xmlText = "'" . $db->escapeString( $xmlText ) . "'";
         return array( 'data_text' => $xmlText );
     }
+
+    protected $deletedStoredObjectAttribute;
```
andrerom: phpdoc: *List of fully deleted object attributes by id, used to know when we don't need to perform additional url cleanup*

andrerom:
+1

yannickroger:
+1

PR #1182, opened 2015-07-06
Fix EZP-24565 AttributeFilter not_in: Escape string only if the value isn't numeric

yannoz:
https://jira.ez.no/browse/EZP-24565

We check if the provided parameter is numeric or not before escaping it. The escaping doesn't rely on the `$filterFieldType` parameter anymore (which isn't set for custom attributes anyway)

### DIFF
--- a/kernel/classes/ezcontentobjecttreenode.php
+++ b/kernel/classes/ezcontentobjecttreenode.php
@@ -1135,6 +1135,7 @@ class eZContentObjectTreeNode extends eZPersistentObject
                         default:
                         {
                             $useAttributeFilter = true;
+                            $filterFieldType = false;
                         } break;
                     }
 
@@ -1288,8 +1289,13 @@ class eZContentObjectTreeNode extends eZPersistentObject
                                     reset( $filter[2] );
                                     while ( list( $key, $value ) = each( $filter[2] ) )
                                     {
+                                      // If filterFieldType is explicitely set
+                                      if( $filterFieldType ){
+                                        $filter[2][$key] = ( $filterFieldType == "integer" ) ? (int) $value : "'" . $db->escapeString( $value ) . "'";
+                                      }else{
                                         // Non-numerics must be escaped to avoid SQL injection
                                         $filter[2][$key] = is_numeric( $value ) ? $value : "'" . $db->escapeString( $value ) . "'";
+                                      }
                                     }
                                     $filterValue = '(' .  implode( ",", $filter[2] ) . ')';
                                 }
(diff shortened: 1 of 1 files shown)


yannickroger:
You patch is breaking the unit test.
If you need to run them locally to troubleshoot, this old documentation works fine: http://www.ezpedia.org/ez/testing_ez_publish_test_system

yannoz:
Hey @yannickroger,

Thanks, I will have a look at it

guillaumelecerf:
What is exactly the problem you're trying to address? IMHO youl need to add a test showing the bug.

yannoz:
Hi @guillaumelecerf ,

There is an example in the bug report, not sure if this is what you expect:
```
eZContentObjectTreeNode::subTreeByNodeId( array(
    'AttributeFilter' => array( array( 'fruit/name', 'not_in', array( 'orange', 'banana' ) ),
    'ClassFilterType' => 'include',
    'ClassFilterArray' => array( 'fruit' )
    ), $parent_node_id );
```
Using the "not_in" operator with custom attributes doesn't work for strings, just for integers.

guillaumelecerf:
@yannoz: isn't this enough: https://github.com/guillaumelecerf/ezpublish-legacy/commit/302b6c8e7ae993feefb239d88ce15b6d6ffa556d ?

yannoz:
@guillaumelecerf Thanks for your reply.

I might be wrong, but i'm afraid it wouldn' change anything : [`$filterFieldType`'s value is 'integer' by default](/guillaumelecerf/ezpublish-legacy/blob/302b6c8e7ae993feefb239d88ce15b6d6ffa556d/kernel/classes/ezcontentobjecttreenode.php#L1020) and [it doesn't get overriden](https://github.com/guillaumelecerf/ezpublish-legacy/blob/302b6c8e7ae993feefb239d88ce15b6d6ffa556d/kernel/classes/ezcontentobjecttreenode.php#L1135) since we use custom attributes.

If I could make a short disgression, I couldn't help noticing you are using `switch` when I use ternary operator instead. Is it recommanded to use the `switch` (much more readable) over the ternary (slightly more efficient) ?

guillaumelecerf:
@yannoz: my bad, should be fixed in https://github.com/guillaumelecerf/ezpublish-legacy/commit/08aebbaab51cc16d4c7105db1d278684331b7c90

About your digression, I tend to always minimize the diff to ease the review ;)

yannoz:
@guillaumelecerf : Yep, it seems perfect to me.

How to you want to proceed next ?

guillaumelecerf:
I created another PR https://github.com/ezsystems/ezpublish-legacy/pull/1187 which replaces this one, that IMHO can now be closed.

Thanks for your time.

yannoz:
Allright, thank you very much !

PR #1185, opened 2015-07-10
Fix EZP-24591: Add quotes in index_cluster ETag value

Opencontent:


andrerom:
+1

yannickroger:
+1
Thanks @Opencontent

PR #1187, opened 2015-07-27
 Fix EZP-24565: correctly quote attributes of custom AttributeFilters

guillaumelecerf:
https://jira.ez.no/browse/EZP-24565

Replaces https://github.com/ezsystems/ezpublish-legacy/pull/1182 : the approach is not right as every items of the in/not_in statement must have the same type (either string or integer) to be properly quoted in the final SQL query.

TODO:
- [x] add tests

### DIFF
--- a/kernel/classes/ezcontentobjecttreenode.php
+++ b/kernel/classes/ezcontentobjecttreenode.php
@@ -1178,6 +1178,7 @@ class eZContentObjectTreeNode extends eZPersistentObject
                                 if ( $filterDataType == 'string' )
                                 {
                                     $sortKey = 'sort_key_string';
+                                    $filterFieldType = 'string';
                                 }
                                 else
                                 {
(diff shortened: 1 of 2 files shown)


guillaumelecerf:
@yannoz: could you please test this version of the fix please ?

xc:
Good one, the issue make the attribute filter: 'in', array( <string> ) not working. 

Tested ok and applied in our project. :)

guillaumelecerf:
@yannoz : ping

guillaumelecerf:
@yannickroger : even if @yannoz did not confirm that it fixes his problem, @xc confirmed it. So I think it's ok to merge this PR.

bdunogier:
Looks good to me. +1, but ideally with a confirmation de from @yannoz that it fixes the issue.

yannickroger:
+1

yannickroger:
Thanks @guillaumelecerf

guillaumelecerf:
Thanks.

PR #1189, opened 2015-07-31
Fix EZP-24641: Copy button next to Published article not triggering an action

yannickroger:
Link: https://jira.ez.no/browse/EZP-24641

## Description
Forms with the same action were nested (see opening tag line 49), which is not allowed by (modern) browsers.
The fix removes the nested form tags.

## Tests
Manual test and sanity checks on the page


glye:
+1

lolautruche:
+1

andrerom:
+1

PR #1190, opened 2015-08-04
Fix EZP-24630: Checkbox value lost when conflicts

yannickroger:
Link: https://jira.ez.no/browse/EZP-24630

## Description
A 10 year old bug (at least).

I added comments in the code with explanations. 
The main idea is that DataTypes check if a value is present in POST and return false when a content should not be modified (nothing to do). 

The problem is that HTTP checkboxes "unchecked" value is "not being present in POST" making it impossible to know if the DataType needs to return false (nothing to do) or to update the data with "unchecked". The bug is that it was updating the data with "unchecked".

In order to fix that, I added checkboxes values in an hidden field in the conflict page which is a "nothing to do" case DataTypewise.

## Tests
Manual tests and sanity tests on checkbox editing

### DIFF
--- a/design/admin/templates/content/edit_conflict.tpl
+++ b/design/admin/templates/content/edit_conflict.tpl
@@ -84,6 +84,22 @@
 <input class="button" type="submit" name="ShowPublishedData" value="{'Show the published data'|i18n( 'design/admin/content/edit_draft' )}" title="{'Create a new draft. The contents of the new draft will be copied from the published version.'|i18n( 'design/admin/content/edit_draft' )}" />
 <input type="hidden" name="PublishAfterConflict" value="1" />
 
+{*
+   EZP-24630: ezboolean work around.
+   Since unchecked textboxes are not present in POST, storing their value in a hidden field.
+ *}
+{foreach $object.contentobject_attributes as $attribute}
+    {if eq( $attribute.data_type_string, 'ezboolean' )}
+        <input
+            id="ezcoa-{$attribute.contentclassattribute_id}_{$attribute.contentclass_attribute_identifier}"
+            class="ezcc-{$attribute.object.content_class.identifier} ezcca-{$attribute.object.content_class.identifier}_{$attribute.contentclass_attribute_identifier}"
+            type="hidden"
+            name="ContentObjectAttribute_data_boolean_{$attribute.id}"
+            value="{$attribute.data_int|wash}"
+        />
+    {/if}
+{/foreach}
+
 </div>
 {* DESIGN: Control bar END *}</div></div>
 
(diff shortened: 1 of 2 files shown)


yannickroger:
ping @andrerom @glye @pspanja

gggeek:
Lovely to see such old mess still being fixed :-). Wish eZP had been built with less reliance on html in the first place...

Btw, on a tangential note: in the new content engine, do all data types properly support (and have documented) nullability/must-have-value? E.g.
- for ints/floats, No-value != 0
- for bools/checkbox, the logic should be ternary (yes/no/not-set)

andrerom:
Not really the approach I had in mind when we discussed it this morning.

What I thought we agreed on was adding a hidden field in the editing template itself, it's value can contain string true/false for what the value was before user potentially edited it. This way you can in the datatype always expect this to be present, if not do nothing. It's a small BC break that needs to be documented, and you'll need to hunt down any editing template in legacy and extensions. However it's not a hack like this.

yannickroger:
@andrerom Sorry this morning I thought we were both talking about this template and the BC break being for people who customized the edit_conflict.tpl.

Now that I understand you suggestion, indeed it has a lot of BC aspects for people with custom template, also that means modifying the behavior of the DataType.

You are sure it is worth it ? This is a 10 year old edge case. And I don't think it's a hack, but really a work around on of our old framework limitation and is really self contained to avoid bad surprises. And so far this is the only template affected.

If you dislike the patch, I can try approach you are mentioning.

andrerom:
I dislike it, but being pragmatic this is ok for now. +1

design/admin/templates/content/edit_conflict.tpl
```diff
@@ -84,6 +84,22 @@
 <input class="button" type="submit" name="ShowPublishedData" value="{'Show the published data'|i18n( 'design/admin/content/edit_draft' )}" title="{'Create a new draft. The contents of the new draft will be copied from the published version.'|i18n( 'design/admin/content/edit_draft' )}" />
 <input type="hidden" name="PublishAfterConflict" value="1" />
 
+{*
+   EZP-24630: ezboolean work around.
+   Since unchecked textbox are not present in POST, storing their value in an hidden field.
```
glye: nitpick: "a hidden field" not "an"

kernel/classes/datatypes/ezboolean/ezbooleantype.php
```diff
@@ -111,6 +111,11 @@ function fetchObjectAttributeHTTPInput( $http, $base, $contentObjectAttribute )
         }
         else
         {
+            /*
+             * Note: in other DataTypes, false is returned if the variable is not present.
+             * Since HTTP checkboxes are not present in the POST when they are not checked,
+             * eZBooleanType might behave a bit differently. See EZP-26630 for a concrete use case.
```
glye: Hawk-eye nitpick: EZP-24630

glye:
+1 besides nitpicks

PR #1191, opened 2015-08-11
EZP-24702: Allow Setup Wizard to work non interactively with kickstart data

andrerom:
https://jira.ez.no/browse/EZP-24702

Needed for @vidarl's work on ezpublish-docker project to simplify it down to just use the code as is and not workaround issues.

### DIFF
--- a/kernel/setup/steps/ezstep_welcome.php
+++ b/kernel/setup/steps/ezstep_welcome.php
@@ -64,6 +64,11 @@ class eZStepWelcome extends eZStepInstaller
         $this->PersistenceList['tests_run'] = $testsRun;
         $this->PersistenceList['optional_tests_run'] = $testsRun;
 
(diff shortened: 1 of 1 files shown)


andrerom:
review ping @crevillo @vidarl @glye

crevillo:
I don't remember what was the problem here. Patch looks good Anyway. I'll let the others review here. +0.7 on my side.

vidarl:
The patch makes it possible to set "continue=true" for the welcome page.
Without it, the user has to click "Next" on the welcome page before the kickstart.ini file "kicks in"....

glye:
+1

andrerom:
@vidarl Give me heads up for when I can merge this (given it will cause conflicts for the patch you use for this).

andrerom:
@vidarl merging.

PR #1192, opened 2015-08-14
Fix EZP-24692: creating content leaves stale video/binary files

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-24692

When creating a video content the first upload creates a binaryfile entry, but saving or publishing the 1st version will create yet another file.
The DB entry in ezbinaryfile is updated for that version, so the file kept in var/storage becomes 'stale'.

This fix makes sure that the old file is removed when storing a new file for a given version.
Notes: 
*  fileName is not the actual (original) filename, but the pseudo-random hash.ext that can be seen under var/storage/
* `deleteStoredObjectAttribute()` method should only remove the file if no longer in use.

### DIFF
--- a/kernel/classes/datatypes/ezbinaryfile/ezbinaryfiletype.php
+++ b/kernel/classes/datatypes/ezbinaryfile/ezbinaryfiletype.php
@@ -293,52 +293,63 @@ class eZBinaryFileType extends eZDataType
         if ( !eZHTTPFile::canFetch( $base . "_data_binaryfilename_" . $contentObjectAttribute->attribute( "id" ) ) )
             return false;
 
-        $binaryFile = eZHTTPFile::fetch( $base . "_data_binaryfilename_" . $contentObjectAttribute->attribute( "id" ) );
+        $httpBinaryFile = eZHTTPFile::fetch( $base . "_data_binaryfilename_" . $contentObjectAttribute->attribute( "id" ) );
 
-        $contentObjectAttribute->setContent( $binaryFile );
+        $contentObjectAttribute->setContent( $httpBinaryFile );
 
-        if ( $binaryFile instanceof eZHTTPFile )
+        if ( $httpBinaryFile instanceof eZHTTPFile )
         {
             $contentObjectAttributeID = $contentObjectAttribute->attribute( "id" );
             $version = $contentObjectAttribute->attribute( "version" );
 
             /*
             $mimeObj = new  eZMimeType();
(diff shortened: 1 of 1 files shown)


andrerom:
+ 0.5 Unsure about this part, maybe @glye or @pspanja have some insight, otherwise it will have to wait until a few of the others are back.

glye:
0.5 as well, don't have much insight here. But if you have verified what deleteStoredObjectAttribute() does, it should be no problem.

yannickroger:
Looks good to me +1.
I would really appreciate @bdunogier 's feedback (sorry) on this one

joaoinacio:
Agreed; a few notes:
`$binaryFile` is an instance of `eZHTTPFile` (the file being uploaded)
`$binary` should be an instance of `eZBinaryFile` - an existing file for the content/version.
The check before file deletion can be seen at [ezbinaryfiletype.php#L204](https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/classes/datatypes/ezbinaryfile/ezbinaryfiletype.php#L204), which looks appropriate.

joaoinacio:
ping @bdunogier

kernel/classes/datatypes/ezbinaryfile/ezbinaryfiletype.php
```diff
@@ -326,7 +326,19 @@ function fetchObjectAttributeHTTPInput( $http, $base, $contentObjectAttribute )
 
             $binary = eZBinaryFile::fetch( $contentObjectAttributeID, $version );
             if ( $binary === null )
+            {
                 $binary = eZBinaryFile::create( $contentObjectAttributeID, $version );
+            }
+            else
+            {
+                // if storing a different file for the same version, see if the existing file can be removed.
+                $fileName = $binary->Filename;
```
bdunogier: I wouldn't bother creating a temp var here.

bdunogier:
The code is really confusing, even if it was before you arrived. I'd like it better if we tried to distinguish `$binary` and `$binaryFile` a bit better in terms of naming.

That done, I'll be +1.

joaoinacio:
@bdunogier: Indeed it is confusing (as noted by my comment above).
I'll see what can be modified without causing even more confusion due to refactoring.

yannickroger:
I think @bdunogier wanted your 2 commits to be squashed. And that would make this PR ready to be merged

joaoinacio:
Got it, thanks - squashed.

yannickroger:
You got your +1's you can merge it

PR #1194, opened 2015-08-23
Allow to hide custom tags

pkamps:
Here an example for the ini file:

[CustomTagSettings]
IsHidden[]
IsHidden[factbox]=true


pkamps:
It's just hiding the value in the dropdown when editors create a custom tag.

peterkeung:
You should create an accompanying issue at jira.ez.no to explain the use case and for better tracking.

Also, consider the case where the custom tag already exists in content: a user might see it in the rich text editor and try to edit it. There should either be some messaging to indicate that it is deprecated / no longer configurable / read-only, or it should allow them to configure it in that case.

pkamps:
I thought about it and I decided that I don't like my approach. I think it's better to clean-up un-used custom tags in the exzml text instead. So let me close this pull request.

PR #1195, opened 2015-08-30
EZP-24744 - Increase password security

pedroresende:
## EZP-24744 - Increase password security

Currently eZ Publish is only using plain text or MD5 for password hash, this improvement implements the usage of BCRYPT to improve the security of the stored passwords

Using https://github.com/ircmaxell/password_compat

### DIFF
--- a/kernel/classes/datatypes/ezuser/ezuser.php
+++ b/kernel/classes/datatypes/ezuser/ezuser.php
@@ -27,6 +27,8 @@ class eZUser extends eZPersistentObject
     const PASSWORD_HASH_MYSQL = 4;
     /// Passwords in plaintext, should not be used for real sites
     const PASSWORD_HASH_PLAINTEXT = 5;
+    /// Passwords in bcrypt of password
+    const PASSWORD_BCRYPT = 6;
 
     /**
      * Max length allowed for a login or a password
@@ -135,6 +137,10 @@ class eZUser extends eZPersistentObject
             {
                 return 'plaintext';
             } break;
+            case self::PASSWORD_BCRYPT:
+            {
+                return 'bcrypt';
+            } break;
         }
     }
 
@@ -165,6 +171,10 @@ class eZUser extends eZPersistentObject
             case 'plaintext':
             {
                 return self::PASSWORD_HASH_PLAINTEXT;
+            } break;
+            case 'bcrypt':
+            {
+                return self::PASSWORD_BCRYPT;
             } break;
         }
     }
@@ -611,6 +621,8 @@ WHERE user_id = '" . $userID . "' AND
             return self::PASSWORD_HASH_MD5_USER;
         else if ( $type == 'plaintext' )
             return self::PASSWORD_HASH_PLAINTEXT;
+        else if ( $type == 'bcrypt' )
+            return self::PASSWORD_BCRYPT;
         else
             return self::PASSWORD_HASH_MD5_PASSWORD;
     }
@@ -1802,6 +1814,19 @@ WHERE user_id = '" . $userID . "' AND
         }
         else if ( $type == self::PASSWORD_HASH_PLAINTEXT )
         {
+            $str = $password;
+        }
+        else if ( $type == self::PASSWORD_BCRYPT )
+        {
+            if ( $hash ) {
+                if ( password_verify( $password, $hash ) ) {
+                    return $hash;
+                } else {
+                    return false;
+                }
+            } else {
+                $str = password_hash( $password, PASSWORD_BCRYPT );
+            }
             $str = $password;
         }
         else // self::PASSWORD_HASH_MD5_PASSWORD
--- a/settings/site.ini
+++ b/settings/site.ini
@@ -563,6 +563,7 @@ UserCreatorID=14
 # md5_site generates password hash from site, user and password
 # plaintext does not generate a hash but has the password as it is, this is not
 # recommended since it is a huge security risc.
+# bcrypt generates password bcrypt from password only.
 # note: password hashes generated with md5_site will not work after
 #       changing the site name.
 HashType=md5_user
--- a/share/db_schema.dba
+++ b/share/db_schema.dba
@@ -7272,7 +7272,7 @@ $schema = array (
       ),
       'password_hash' => 
       array (
-        'length' => 50,
+        'length' => 255,
         'type' => 'varchar',
         'default' => NULL,
       ),
--- a/update/common/scripts/5.4/updatepasswordhashlength.php
+++ b/update/common/scripts/5.4/updatepasswordhashlength.php
@@ -1,0 +1,27 @@
+<?php
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ * @package kernel
+ */
+
+require 'autoload.php';
+
+set_time_limit( 0 );
+
+$cli = eZCLI::instance();
+$script = eZScript::instance(
+    array(
+        'description' => 'Script responsible for changing the number of characters the password_hash column, in ezuser table',
+    )
+);
+
+$script->initialize();
+$script->startup();
+
+$db = eZDB::instance();
+
+$rows = $db->query( "ALTER TABLE `ezuser` CHANGE COLUMN `password_hash` `password_hash` VARCHAR(255) NULL DEFAULT NULL ;" );
+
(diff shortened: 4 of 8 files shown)


share/db_schema.dba
```diff
@@ -7272,7 +7272,7 @@ $schema = array (
       ),
       'password_hash' => 
       array (
-        'length' => 50,
+        'length' => 255,
```
yannickroger: If you update the schema you also need a BC doc and an update script.

kernel/classes/datatypes/ezuser/ezuser.php
```diff
@@ -27,6 +27,8 @@ class eZUser extends eZPersistentObject
     const PASSWORD_HASH_MYSQL = 4;
     /// Passwords in plaintext, should not be used for real sites
     const PASSWORD_HASH_PLAINTEXT = 5;
+    /// Passwords in bcrypt of password
+    const PASSWORD_BCRYPT = 6;
```
michalpipa: I wouldn't call it BCRYPT, because BCRYPT is default algorithm in current `password_hash()` implementation and can change in the future.

Maybe `PASSWORD_HASH` would be a better name.

pedroresende: @michalpipa since password_hash is only present on PHP 5.5, I'm currently using https://github.com/ircmaxell/password_compat has you can see in the composer.json

andrerom: @michalpipa Here @pedroresende is explicitly using php's PASSWORD_BCRYPT constant and not PASSWORD_DEFAULT. I was thinking of suggesting to change that, but I think I prefer this as otherwise users using different PHP versions in production and dev might get into situation where their database contains future hashes not supported by given PHP version.

So it's ok that he specifies as long as we are specific to `password_hash` as well.

michalpipa: > Here @pedroresende is explicitly using php's PASSWORD_BCRYPT constant and not PASSWORD_DEFAULT.

Ok.

update/common/scripts/5.4/updatepasswordhashlength.php
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+require 'autoload.php';
+
+set_time_limit( 0 );
+
+$cli = eZCLI::instance();
+$script = eZScript::instance(
+    array(
+        'description' => 'Script responsible for chaging the number of character the password_hash column, from ezuser table',
```
joaoinacio: `s/chaging/changing/`
`s/character/characters/`

pedroresende: @joaoinacio thanks, fixed, rebased and pushed

settings/site.ini
```diff
@@ -562,6 +562,7 @@ UserCreatorID=14
 # md5_user generates password hash from user and password.
 # md5_site generates password hash from site, user and password
 # plaintext does not generate a hash but has the password as it is, this is not
+# bcrypt generates password bcrypt from password only.
```
joaoinacio: Should be added on the line bellow, as the comment above continues to next line ( '... is not recommented ... ')

pedroresende: @joaoinacio thanks, fixed, rebased and pushed

update/database/mysql/5.9/dbupdate-5.4.0-to-5.90.0.sql
```diff
@@ -0,0 +1,2 @@
+-- See https://jira.ez.no/browse/EZP-24744 - Increase password security
+ALTER TABLE ezuser CHANGE password_hash password_hash VARCHAR(255) NULL DEFAULT NULL;
```
yannickroger: Nitpick, add end of line (same in the other sql script)

yannickroger:
+1

update/database/mysql/5.9/dbupdate-5.4.0-to-5.90.0.sql
pedroresende: @yannickroger added, rebased and pushed

settings/site.ini
```diff
@@ -563,6 +563,7 @@ UserCreatorID=14
 # md5_site generates password hash from site, user and password
 # plaintext does not generate a hash but has the password as it is, this is not
 # recommended since it is a huge security risc.
+# bcrypt generates password bcrypt from password only.
 # note: password hashes generated with md5_site will not work after
 #       changing the site name.
 HashType=md5_user
```
emodric: So why not change the default hash type to bcrypt, like in https://github.com/ezsystems/ezpublish-kernel/pull/1388 ?

pedroresende: @andrerom ping

andrerom: why no, we where thinking bc.

emodric: @andrerom Nice! :+1:

kernel/classes/datatypes/ezuser/ezuser.php
```diff
@@ -1804,6 +1816,19 @@ static function createHash( $user, $password, $site, $type, $hash = false )
...
+            if ( $hash ) {
+                if ( password_verify( $password, $hash ) ) {
+                    return $hash;
+                } else {
+                    return false;
+                }
+            } else {
+                $str = password_hash( $password, PASSWORD_BCRYPT );
+            }
+            $str = $password;
```
emodric: This here is wrong. This means that unhashed password will be returned from this method, always.

settings/site.ini
emodric: So, changing this to `bcrypt`?

update/common/scripts/5.4/updatepasswordhashlength.php
```diff
@@ -0,0 +1,27 @@
+<?php
```
emodric: What's the use of the script if we have upgrade SQL scripts?

kernel/classes/datatypes/ezuser/ezuser.php
emodric: Why not name it `PASSWORD_HASH_BCRYPT` to make it consistent with other constants?

kernel/classes/datatypes/ezuser/ezuser.php
```diff
@@ -1804,6 +1816,19 @@ static function createHash( $user, $password, $site, $type, $hash = false )
...
         }
+        else if ( $type == self::PASSWORD_BCRYPT )
+        {
+            if ( $hash ) {
+                if ( password_verify( $password, $hash ) ) {
+                    return $hash;
+                } else {
+                    return false;
+                }
+            } else {
```
crevillo: nitpick: CS. i know there are changes about this in ez5, but with legacy we are still applying old coding standards, right? If so, else should be in new line as you do in https://github.com/ezsystems/ezpublish-legacy/pull/1195/files#diff-9aabf5ca51952bbd469991c839db4f3eR1819

update/common/scripts/5.4/updatepasswordhashlength.php
crevillo: apart from @emodric comment, i'm a bit confused. why this is script is in ```/5.4``` but the sql is in ```/5.9```?

emodric:
Default hashes in `db_data.dba` and `cleandata.sql` should be changed to bcrypt...

andrerom:
Remaing changes seems to be here https://github.com/netgen/ezpublish-legacy/commit/d630485600f4fa7b3fded21316e3aae41636bfb9 So that could be opened as new PR for this.

Given this feature has not been merged into 6.11, I'll de-scope it from 2017.08. We can aim for this being part of 2017.10 _(1.12)_ instead.

andrerom:
> Remaining changes seems to be here https://github.com/netgen/ezpublish-legacy/commit/d630485600f4fa7b3fded21316e3aae41636bfb9 So that could be opened as new PR for this.

Actually, with PHP 7.2 introducing new stronger algorithm that probably will become default in the future, it would be better to just use default here. So constant name can for instance be `PASSWORD_HASH_PHP` instead of `PASSWORD_HASH_BCRYPT`

/cc @glye

update/common/scripts/5.4/updatepasswordhashlength.php
andrerom: I guess we don't need the script, the sql however should be updated to reflect the release it will be in 2017.10 _(if it makes it for that)_.

emodric:
Closed with #1322 .

PR #1196, opened 2015-09-01
EZP-24756: ezimage datatype - do not store file if not valid

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-24756
#### Problem:

When uploading invalid files to an ezimage attribute (empty files, non-images, etc), the input is validated but the file is still stored and eZ still tries to generate image aliases.
#### Solution:

This modifies the ezimage datatype so that the file is not stored if the validation result is `STATE_INVALID`, with a minor refactor.

### DIFF
--- a/kernel/classes/datatypes/ezimage/ezimagetype.php
+++ b/kernel/classes/datatypes/ezimage/ezimagetype.php
@@ -295,8 +295,8 @@ class eZImageType extends eZDataType
 
         $content = $contentObjectAttribute->attribute( 'content' );
         $httpFileName = $base . "_data_imagename_" . $contentObjectAttribute->attribute( "id" );
-
-        if ( eZHTTPFile::canFetch( $httpFileName ) )
+        if ( $contentObjectAttribute->IsValid != eZInputValidator::STATE_INVALID
+            && eZHTTPFile::canFetch( $httpFileName ) )
         {
             $httpFile = eZHTTPFile::fetch( $httpFileName );
             if ( $httpFile )


andrerom:
@joaoinacio .. would it be possible to just show the fix? why did you refactor this? it is quite hard to review when you change everything in the function, and besides this is legacy, so you need pretty good reasons to refactor things here *(no tests)*.

joaoinacio:
@andrerom: Yes there is a small refactor but the exact same behavior should be maintained.

The fix is essentially to only store the httpFile in content attribute when previous validation check did not return invalid. ref:
`if ( $contentObjectAttribute->IsValid != eZInputValidator::STATE_INVALID )`

kernel/classes/datatypes/ezimage/ezimagetype.php
```diff
@@ -284,40 +284,32 @@ function validateObjectAttributeHTTPInput( $http, $base, $contentObjectAttribute
...
             $imageAltText = $http->postVariable( $base . "_data_imagealttext_" . $contentObjectAttribute->attribute( "id" ) );
-            $hasImageAltText = true;
+            $content->setAttribute( 'alternative_text', $imageAltText );
         }
 
-        $content = $contentObjectAttribute->attribute( 'content' );
-        $httpFileName = $base . "_data_imagename_" . $contentObjectAttribute->attribute( "id" );
-
-        if ( eZHTTPFile::canFetch( $httpFileName ) )
+        if ( $contentObjectAttribute->IsValid != eZInputValidator::STATE_INVALID )
```
joaoinacio: The condition could be reversed here instead (` == STATE_ACCEPTED`)

joaoinacio:
@andrerom: split in separate commits, I guess the cleanup could be omitted if necessary.

andrerom:
+1 w/o cleanup, cleanup adds somewhat risk, and obfuscate the git history any time in the future when we need to do a blame. If we planned to make a new release we could have considered the cleanup for master, but we don't.

andrerom:
And thanks for the split, was much easier to do a review now and the change makes perfect sense. Lest hope that one alone does not cause to many regressions for multi upload or what not :)

PR #1197, opened 2015-09-01
EZP-20547: Allow specifying max. connect timeout for link check

joaoinacio:
This implements JIRA https://jira.ez.no/browse/EZP-20547, also resolves https://jira.ez.no/browse/EZP-24738


andrerom:
+1

pedroresende:
+1

yannickroger:
+1 but I would keep `2` for a perfect BC

andrerom:
No, do 3. If the value is to low, it is to low :)

PR #1199, opened 2015-09-02
customer order view workaround for "+" in email addresses

pbek:
I added a workaround to the get the customer order view working with email addresses that include a "+" character.

First I thought only an "urlencode" was missing, but then I realized it wasn't possible to get a "+" across $Params.

### DIFF
--- a/cronjobs/linkcheck.php
+++ b/cronjobs/linkcheck.php
@@ -79,7 +79,7 @@ foreach ( $linkList as $link )
               // Check if it is a valid internal link.
               foreach ( $siteURLs as $siteURL )
               {
-                  $siteURL = preg_replace("/\/$/e", "", $siteURL );
+                  $siteURL = preg_replace("/\/$/", "", $siteURL );
                   $fp = @fopen( $siteURL . "/". $url, "r" );
                   if ( !$fp )
                   {
--- a/design/admin/templates/setup/datatype_code.tpl
+++ b/design/admin/templates/setup/datatype_code.tpl
@@ -26,7 +26,7 @@ class {$full_class_name} extends eZDataType
     */
     function {$full_class_name}()
     {literal}{{/literal}
-        $this->eZDataType( self::{$constant_name}, "{$desc_name}" );
+        parent::__construct( self::{$constant_name}, "{$desc_name}" );
     {literal}}{/literal}
 
 {if $class_input}
--- a/design/standard/templates/setup/datatype_code.tpl
+++ b/design/standard/templates/setup/datatype_code.tpl
@@ -26,7 +26,7 @@ class {$full_class_name} extends eZDataType
     */
     function {$full_class_name}()
     {literal}{{/literal}
-        $this->eZDataType( self::{$constant_name}, "{$desc_name}" );
+        parent::__construct( self::{$constant_name}, "{$desc_name}" );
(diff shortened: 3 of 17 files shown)


andrerom:
+1

kernel/shop/customerorderview.php
```diff
@@ -16,6 +16,13 @@
 $tpl = eZTemplate::factory();
 
 $Email = urldecode( $Email );
+
+// workaround because it seems not possible to get an urlencoded "+" character accross $Params
+if ( $http->hasGetVariable( "email" ) )
+{
+    $Email = $http->getVariable( "email" );
+}
+
```
crevillo: But, if i'm not wrong, here you are remove Email Param from the url, right? i mean, [this](https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/shop/module.php#L112) is not valid as you won't have Email as Param anymore

pbek: I use a normal "get" parameter in orderlist.tpl, because "+" get through with them. I give this parameter a higher priority in the customerorderview.php. Does that seem wrong to you, @crevillo ?

Of course fixing the bug (?) with $Params would be the better solution.

crevillo: @pbek. no, the patch is good. forget my comment. just though that now [this line](https://github.com/pbek/ezpublish-legacy/blob/master/kernel/shop/customerorderview.php#L10)
could throw a php notice or something now $Params['Email'] won't be set in this new urls.

pbek: I'm not very happy with that patch, because the real problem lies in $Params (and a missing `urlencode()`). ;)

pbek:
I created an issue now, because I get no reaction on the pull request...
https://jira.ez.no/browse/COM-19862

SylvainGuittard:
Moved to eZ Publish Project (EZP): https://jira.ez.no/browse/EZP-24798

kernel/shop/customerorderview.php
andrerom: What do you have in mind changing it to?

as for Params, it kind of still need to be there for bc.

pbek: I think I fail to understand what you are talking about. :)
Changing what to? And what "bc"?

pbek:
any news here?

andrerom:
@pbek might want to move ca0268d into own branch against master ;)

pbek:
@andrerom I just realised that if you push changes to your own fork an open pull request gets updated automatically... Since I want to create a pull request for fixing PHP 7.2 deprecations warning/errors I had to remove my fork, since you cannot create a new one...

andrerom:
hmm, should be possible, you just need to use different branches.

pbek:
ah, that sounds right. so I could have created another branch from `master` and work with that one, I will try that for my PHP 7.2 pull request

pbek:
Yes, thank you. I guess it's best always to create a branch for a new pull request!

pbek:
Do you mean my locally checked master branch or the remote master branch (of my fork)?

pbek:
I meanwhile created https://github.com/ezsystems/ezpublish-legacy/pull/1390 and https://github.com/ezsystems/ezpublish-legacy/pull/1391. ;)

PR #1200, opened 2015-09-10
checking if object is already created

pbek:
checking if object is already created before hiding the login text input, otherwise the login can't be changed (even if it is empty and throws an error) after the first form submit if the content object was newly created (example: user registration)

### DIFF
--- a/design/standard/templates/content/datatype/edit/ezuser.tpl
+++ b/design/standard/templates/content/datatype/edit/ezuser.tpl
@@ -18,7 +18,7 @@
 {* Username. *}
 <div class="element">
     <label for="{$id_base}_login">{'Username'|i18n( 'design/standard/content/datatype' )}:</label>
-    {if $attribute.content.has_stored_login}
+    {if and( $attribute.content.has_stored_login, $attribute.object.published )}
         <input id="{$id_base}_login" type="text" name="{$attribute_base}_data_user_login_{$attribute.id}_stored_login" size="16" value="{$attribute.content.login|wash()}" disabled="disabled" />
         <input id="{$id_base}_login_hidden" type="hidden" name="{$attribute_base}_data_user_login_{$attribute.id}" value="{$attribute.content.login|wash()}" />
     {else}


yannickroger:
Thanks for your patch.
Can you create an issue corresponding to your patch on http://jira.ez.no/
Make sure to add detailed steps to reproduce the issue.

pbek:
done: https://jira.ez.no/browse/COM-19861

SylvainGuittard:
Thanks @pbek 
Issue moved to EZP Project: https://jira.ez.no/browse/EZP-24797

yannickroger:
@pbek In the Jira issue, you mean to press the "Save Draft" button right ?

pbek:
@yannickroger, thank you. seems like I could not select "eZ Publish Platform" :/
Then my other issue has to be moved too:
https://jira.ez.no/browse/COM-19862
TIA

pbek:
@yannickroger, you can press "publish" as well, because you will get an error message, because no email address (and login name) was entered. ;)

yannickroger:
@pbek Got it. I updated the discription of the issue.

+1

pbek:
@yannickroger, it's "a bit" of a show stopper as well in user registration, because a user cannot complete the registration process unless he/she presses "remove draft" and starts all over. ;)
Especially annoying if you have the RegExp check for the login name on, because you cannot enter a correct login name after you get the error message.
I wonder how this bug could exists for a decade. :)

yannickroger:
Actually "draft" (they are not real drafts as users are not versioned) for users have been added a couple years ago :) So this bug could be a bit more recent.

pbek:
I just checked a eZ Publish 3.7.3 site, it doesn't have this bug. :)
And neither has eZ Publish Community Project 2013.7.
You are right!

guillaumelecerf:
Isn't the username taken into account to calculate the password hash ?
If so, the stored hash won't be valid if you allow the user to change its username.

pbek:
As long the account isn't created the username should be changeable, otherwise you can't enter it at all if you press the publish button before you enter it (and you can't correct your username if an error check failed). :)

yannickroger:
@pbek did you test @guillaumelecerf 's use case with you patch ?

lolautruche:
@guillaumelecerf might be right, we need to check this.

pbek:
What exactly is the use case we want to test, @guillaumelecerf and @yannickroger ?
The patch merely checks if the content object is published, if it was published once you can't edit your login name again.

yannickroger:
ping @guillaumelecerf can you provide a step by step scenario of the use case you mentioned please ?

guillaumelecerf:
I'm rather busy ATM, I will try to find time by the end of the week.

pbek:
any news, @guillaumelecerf ?

guillaumelecerf:
Sorry for the delay :(

If you look at https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/classes/datatypes/ezuser/ezuser.php#L1780 you see that the password hash is generated from the username (amongst other things).

Maybe you could try to modify the login of a user and see if you're still able to login, but I doubt you will.

pbek:
Thank you for your answer @guillaumelecerf ! Can you please describe the steps I should test.
You can try to register at <https://www.astrosofa.com/user/register>. My patch is in place there.

PR #1201, opened 2015-09-12
Fix EZP-24800: Support "allowed classes" limitation for object relation fields

peterkeung:
https://jira.ez.no/browse/EZP-24800

This PR ports the code over from the object relation list datatype.

Since the object relation datatype stored its various other parameters (start node and selection method) in other database fields, for backwards compatibility only the "allowed classes" list is stored in XML here.


peterkeung:
Ping @bgamrat @benkmugo @ernestob @thiagocamposviana for review

benkmugo:
Would be good to have this back as it allows for tighter editorial control.
+1 from me

thiagocamposviana:
It is a must have feature for ez object relation list, sometimes we get caught using the object relations attribute just because of this feature when an object relation would be overall a better solution if it had such option to filter the list by class.

bgamrat:
+1

dougplant:
Thiago is right.

jeromegamez:
:+1: !

dfritschy:
+1

peterkeung:
Ping @yannickroger @andrerom let's get this in! :tada:

granitegreg:
Thanks, @peterkeung!
+1

alexgorski:
Yes! Please get it in @yannickroger @andrerom @pkamps :speedboat:

angatqs:
It's important to have that ability to limit by class. Otherwise our field-specific searches return a bunch of junk and we cannot limit the related objects that editors can select.

Being one of Mugo's client on the Enterprise subscription, thanks for bringing this to my attention @peterkeung 
+1

bdunogier:
Hello there.

A question for those who added a +1 for this: is it a "+1 I want this !" or "+1 I've read your code / tested your patch, and it works great" ? :-)

andrerom:
Good morning,

the patch looks good, but as we are not really doing new releases of legacy it would make more sense to put this in a extension. Furthermore our plan around this for the future *(which everyone is free to contribute to speed up on)*, is to add a field definition flag on RelationList to add support for single mode. If this where to be contributed to legacy we could consider this together with effort for new stack to keep them aligned on this.

When that is in place we can deprecate objectrelation field/datatype, and start to think about adding system wide location relation support again *(in new stack)*.

Hope this makes sense for you as well, best
André

dfritschy:
@bdunogier  yes, I gave +1 in favour for this feature, looked at the code, but did not test it. 

@andrerom this is a good point and I would favour to concentrate on one datatype. Instead of a *single mode* I would however suggest a *minimum/maximum range* of related objects. This would suit many of our use cases much better.

andrerom:
> minimum/maximum range of related objects

makes sense

benkmugo:
@bdunogier the +1 as in: would like this feature, also reviewed the code

The focus on one data type going forward featuring the min/max range restriction that @dfritschy mentioned sounds like a great idea.

I'm still in favour of getting this patched up in legacy, since a number of clients are still on that.

bgamrat:
My +1 = reviewed the code and consider it and important update

peterkeung:
I strongly disagree about putting this in an extension.  For sure the long-term plan would be better with a single object relation list extension, but we need to consider backwards compatibility.  We had 3 Enterprise clients complain about this over the past month (these sorts of things always seem to happen at the same time, which is great), and this is a drop-in that would 1) Require no big migrations for data and to update all the template code; and 2) Be easily supported as part of the kernel because it follows the object relation list model.

pkamps:
much :heart:

andrerom:
@peterkeung I get you pov, but we need to maintain for a long time, so if it is not the direction we plan to take the features, then it does not fit. As we will create migration scripts and fix bugs on this long after this PR has been done.

peterkeung:
I still see a strong case AND fit for this based on the fact that it copies the object relation list implementation, which mitigates your concerns.
* Long-term maintenance would have to be done to the object relation list implementation (if any) and that is shared and stable code.
* You are going to have to migrate object relation (singular) to that super, single field type anyway, which would need to receive this list of allowed classes the same way object relation list (plural) has it implemented.

My last card to pull is going to be the strong support and need for this from clients who are going to be paying eZ for legacy for at least a few years as we go through this migration process. Forcing us to, in essence, fork the code is going to result in more pain for all parties in the long-run.

andrerom:
Ok, but as ez engineering is in lock down *(scope freeze last week, feature freeze on monday)* on the the Platform/Studio releases the next few months, we need help to complete this so it is mergeable as a feature. That should be doable if we round up the others supporting this feature on slack to organize this and split out tasks, afaik:
- this one
- Migration script in EzPublishMigrationBundle from Relation to RelationList *(singular feature can be added later, however it would be more efficient to do it now as part of all this)*
- Platform kernel Relation field type support (persistence/legacy, fieldtype, ..)
- Platform ui js part content editing limitation, passing this on to UDW *(Universal Discovery Widget)*
- Platform ui admin symfony part content type editing allowing selecting the limitation

What I can offer is a one from engineering pr task here to act as mentor, and we will also deal with the rest (doc, testing, review, support).

peterkeung:
A baby blew out my schedule...

But anyway, that sounds interesting. The challenge will be for someone at Mugo to have enough time to dedicate to what would be a mini-project. Until then we're resigned to having to build a worse solution such as a template override only.

andrerom:
Alternative _(@iosolinf brought this back up on slack)_: If the "multiple fields" option is a content model feature in the future _(asked for at several workshops with partners/community in the past as part of set of relaxed schema features)_. If so it can make sense to do what is here, more or less as is.

But we'll also need it for new stack, so if someone is up for it:
- Add `selectionContentTypes` to [Relation\Type](https://github.com/ezsystems/ezpublish-kernel/tree/master/eZ/Publish/Core/FieldType)
  - Add serialization in [RelationConverter](https://github.com/ezsystems/ezpublish-kernel/tree/master/eZ/Publish/Core/Persistence/Legacy/Content/FieldValue/Converter)* like here
- Copy/Reuse option from Relation list to Relation in [repository-forms](https://github.com/ezsystems/repository-forms/tree/master/lib/FieldType/Mapper)
- Copy/Reuse logic for Platform UI from [relation list](https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/public/js/views/fields/ez-relationlist-editview.js#L194) to [relation](https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/public/js/views/fields/ez-relation-editview.js#L135)


This is less work then what I proposed above.

* @peterkeung ~Do we need to serialize this to xml? given this field type puts the other fields in data_int* columns, we could probably serialize it to json if we want to simplify a bit.~ On the other side if we stick to same format migration between the two is simple.

volkanakb:
After update i get an error, would appreciate if anyone could help. I guess there is need for a update script:
Call to a member function getAttribute() on null in /home/vasd2017/html/shared/vendor/ezsystems/ezpublish-kernel/eZ/Publish/Core/Persistence/Legacy/Content/FieldValue/Converter/RelationConverter.php on line 150

andrerom:
@volkanakb which platform version do you use? This is only compatible with eZ Platform 1.11 and up, which is why we did eb3312ed1e62ad307645c3da1547518a0292eda7 as part of the new [2017.08 release](https://github.com/ezsystems/ezpublish-legacy/milestone/1?closed=1)

volkanakb:
@andrerom thanks for your help. 6.11 and 2017.08 this is the affected row in the ezcontentobject_attribute table:

6583 | 503 | 102 | 0 | 275 |   | ezobjectrelation | 6583 | ger-DE | 5 | 275 |   | 31

andrerom:
@volkanakb can you post line 150 +/1 a line from home/vasd2017/html/shared/vendor/ezsystems/ezpublish-kernel/eZ/Publish/Core/Persistence/Legacy/Content/FieldValue/Converter/RelationConverter.php  ?

volkanakb:
@andrerom 
if ($selectionType = $dom->getElementsByTagName('selection_type')) {
           $fieldSettings['selectionMethod'] = (int)$selectionType->item(0)->getAttribute('value');
}

andrerom:
And what do you have in relevant `ezcontentclass_attribute` row?

volkanakb:
Hi @andrerom,
i found the problem finally, as soon as i publish a Contenttype definiton with a objectrelation it writes into the ezcontentclass_attribute following value: 

```
<?xml version="1.0" encoding="utf-8"?>
<related-object><constraints/></related-object>
```

As soon i delete it, it works again. 

`UPDATE `ezcontentclass_attribute` SET `data_text5` = '' WHERE `data_type_string` = 'ezobjectrelation'`

andrerom:
Hi @volkanakb, completely forgot about this, in the mean time I have another user that got into the same issue, and your comment made me realize why this happens. patch on legacy and kernel are not doing the same thing, the one on kernel ports over all the xml features, while this one only some.

I'll care a fix for kernel to avoid the exception ,and make sure dataInt1 and dataInt2 is still written to for better compatibility with this patch as is.

volkanakb:
@andrerom thx, but i found one more typo, there are 3 parentheses instead of 2

https://github.com/ezsystems/ezpublish-legacy/blob/a4a0470f8d80f012fe14e4f8ab11c7d14375986c/design/standard/templates/content/datatype/edit/ezobjectrelation.tpl#L103

andrerom:
@volkanakb Replied on your comment there.
As for kernel patch, please have a look and test/review here: https://github.com/ezsystems/ezpublish-kernel/pull/2115

PR #1203, opened 2015-09-28
Fix EZP-24889: Clear in-memory cache for moved nodes in delayed indexing script

peterkeung:
We clear in-memory cache for each normal indexed object, but not for moved node subtrees.

https://jira.ez.no/browse/EZP-24889


bdunogier:
Seems legit, +1

lolautruche:
+1

PR #1204, opened 2015-09-28
Fix EZP-24890: Delayed indexing should fail gracefully on trashed / broken node

peterkeung:
The diff is misleading -- the fix is simply to mark the item to be removed instead of continuing on with the loop. I also renamed the offset and limit for the "move" section so it doesn't mess up the main offset and limit.

An alternative (that should perhaps be used together with this fix) is to make the trash action remove the "index_moved_node" entry from ezpending_actions, but then you still have a cleanup issue for any sites with this problem.

### DIFF
--- a/cronjobs/indexcontent.php
+++ b/cronjobs/indexcontent.php
@@ -62,48 +62,50 @@ while( true )
                     if ( !( $node instanceof eZContentObjectTreeNode ) )
                     {
                         $cli->error( "An error occured while trying fetching node $nodeId" );
-                        continue;
+                        $removeFromPendingActions = true;
                     }
-
-                    $offset = 0;
-                    $limit = 50;
-
-                    $params = array( 'Limitation' => array(), 'MainNodeOnly' => true );
-
-                    $subtreeCount = $node->subTreeCount( $params );
-
-                    while ( $offset < $subtreeCount )
+                    else
                     {
-                        $subTree = $node->subTree(
-                            array_merge(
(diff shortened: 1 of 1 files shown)


yannickroger:
@peterkeung you need to rebase this branch

peterkeung:
@yannickroger rebase done :)

andrerom:
@peterkeung Is it possible to reduce amount of diff in the patch? it is close to unreadable, and will be prone for backport conflicts.

For instance *(and please look out for dinosaurs jumping on you when trying this)*, but what if you used goto in the if block to avoid having to wrap the rest of the code in a else block?

peterkeung:
Nice suggestion from @bgamrat and @thiagocamposviana was to just duplicate the code, which I've done now: https://github.com/ezsystems/ezpublish-legacy/pull/1206.  I don't like it as much, especially for such a short file, but it's a cleaner diff.

andrerom:
damn, I was looking forward to:

joaoinacio:
A more elegant approach could be to:
```php

                     if ( !( $node instanceof eZContentObjectTreeNode ) )
                     {
                        $cli->error( "An error occured while trying fetching node $nodeId" );
-                       continue;
+                       $removeFromPendingActions = true;
+                       break 2;
                     }
```

peterkeung:
break N seems less clear to me

peterkeung:
Closing as #1206 has been merged

PR #1205, opened 2015-09-29
Save sort order when sort changed

Alexnder:


andrerom:
This is a retrieval function, it should not perform changes on that data.

Alexnder:
But I don't know another way to save sort order.

blankse:
There is an ordering tab in the backend ;)

image

Alexnder:
Oh... Thanx!

PR #1206, opened 2015-09-29
Fix EZP-24890: Delayed indexing should fail gracefully on trashed / broken node (alt)

peterkeung:
An alternative to https://github.com/ezsystems/ezpublish-legacy/pull/1204 with a simpler diff: duplicating the DB commands.

### DIFF
--- a/cronjobs/indexcontent.php
+++ b/cronjobs/indexcontent.php
@@ -62,6 +62,8 @@ while( true )
                     if ( !( $node instanceof eZContentObjectTreeNode ) )
                     {
                         $cli->error( "An error occured while trying fetching node $nodeId" );
+                        $db->query( "DELETE FROM ezpending_actions WHERE action = '$action' AND param = '$objectID'" );
+                        $db->commit();
                         continue;
                     }
 


emodric:
@peterkeung You have pa permission change in your diff (100644 -> 100755). I don't think this is required since this is a cronjob script, executed through runcronjobs.php, not directly, so no need for execute permissions.

cronjobs/indexcontent.php
```diff
@@ -62,22 +62,24 @@
                     if ( !( $node instanceof eZContentObjectTreeNode ) )
                     {
                         $cli->error( "An error occured while trying fetching node $nodeId" );
+                        $db->query( "DELETE FROM ezpending_actions WHERE action = '$action' AND param = '$objectID'" );
+                        $db->commit();
```
joaoinacio: Won't this db `commit()` potentially introduce a transaction counter mismatch somewhere?

peterkeung: continue() brings us into a new foreach loop, which then starts a new transaction with $db->begin(), right?

peterkeung:
@emodric Good catch!  I completely missed that.  I've removed that file permission change now.

andrerom:
Besides the unrelated offset/limit variable renaming which should be removed;  +1 *(also for backport)*

peterkeung:
Variable clash fix has been removed now. We can create a separate ticket for that.

joaoinacio:
+1

yannickroger:
+1
@peterkeung can you do the rename mentioned by @joaoinacio and ping me, I'll merge it asap.

yannickroger:
@peterkeung @joaoinacio I will do the rename right now.

yannickroger:
My bad you did it @peterkeung

yannickroger:
Thanks @peterkeung

yannickroger:
You did not do it… sorry… I added a note of the commit. They are duplicates so it's not too bad anyways.

PR #1208, opened 2015-10-07
Fix EZP-24918: Delayed indexing: moved nodes logic skips normal pending objects

peterkeung:
This was discovered when fixing #1206 and split into its own issue.

Link: https://jira.ez.no/browse/EZP-24918

### DIFF
--- a/cronjobs/indexcontent.php
+++ b/cronjobs/indexcontent.php
@@ -67,19 +67,19 @@ while( true )
                         continue;
                     }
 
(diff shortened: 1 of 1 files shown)


peterkeung:
Ping @emodric @andrerom @joaoinacio @yannickroger who have technically already seen this code.

andrerom:
+1

yannickroger:
+1
Thanks @peterkeung

PR #1209, opened 2015-10-07
Fix EZP-24919: Clear view/content cache when an object gets re-indexed

pkamps:


pkamps:
Closed in favor of https://github.com/ezsystems/ezpublish-legacy/pull/1210

PR #1210, opened 2015-10-07
Fix EZP-24919: Clear view/content cache when an object gets re-indexed

pkamps:
https://jira.ez.no/browse/EZP-24919

### DIFF
--- a/cronjobs/indexcontent.php
+++ b/cronjobs/indexcontent.php
@@ -114,6 +114,7 @@ while( true )
             if ( $removeFromPendingActions )
             {
                 $db->query( "DELETE FROM ezpending_actions WHERE action = '$action' AND param = '$objectID'" );
+                eZContentCacheManager::clearContentCacheIfNeeded( $objectID );
             }
             else
             {


andrerom:
Doesn't really look like something to do on delayed indexing, why is this needed? Is your content view using data from search instead of fetch?

joaoinacio:
My 2 cents:

Given the description on JIRA issue, sounds like there is indeed some content being fetched from SOLR. But what happens if that content/view has no direct relation to the object(s) being indexed?

Cearing caches can be quite expensive performance-wise, indexing many objects could potentially  cause the system to slow to a crawl (it's possible that the CacheTreshold value is ignored here if objects are  cleared one-by-one).

IMO, even if it may be reasonable, the described issue isn't really a bug but rather a potential feature/enhancement for custom implementations. 
However, the suggested approach looks way overkill...

pkamps:
I don't think Mugo is the only partner using SOLR fetches for bigger eZ Publish installations. I agree, it's maybe not affecting all clients but it's probably not uncommon to end up with SORL driven landing/section pages - at least it's not uncommon for Mugo's enterprise clients.

That said, I think we all agree that there is a caching issue for those installations and the only way I see to solve it is to change the order of execution here: re-index object then clear content view cache.

If I would use a new setting and wrap the cache clearing into a if-statement, would that be a pull-request you'd re-consider?

glye:
I sent this to QA, as a fix for https://jira.ez.no/browse/EZP-27460

tharkun:
+1 it would solve a lot of issues we face on pages using solr search

glye:
QA approved & merged. Thanks @pkamps!

PR #1211, opened 2015-10-22
Fix EZP-25017: Administration Interface object title runs over language indicator

peterkeung:
https://jira.ez.no/browse/EZP-25017

This restores the language indicator to be below the title, on the same line as the author, object ID, and node ID information.

### DIFF
--- a/design/admin/stylesheets/pagelayout.css
+++ b/design/admin/stylesheets/pagelayout.css
@@ -875,7 +875,6 @@ div.content-navigation div.context-information p.left
 div.content-navigation div.context-information p.right
 {
     position:absolute;
-    top: -2.5em;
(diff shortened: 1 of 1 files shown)


andrerom:
Screenshot?

peterkeung:
Before (see ticket attachment): https://jira.ez.no/browse/EZP-25017
After: https://dl.dropboxusercontent.com/u/23142/temp/language_after.png

The "after" is actually where the language indicator used to be in older versions of the Administration Interface.

andrerom:
+1

yannickroger:
+1
Thanks @peterkeung

yannickroger:
@andrerom backport ?

PR #1213, opened 2015-11-02
z

bar-low:


composer.json
```diff
@@ -1,6 +1,6 @@
 {
     "name":        "ezsystems/ezpublish-legacy",
-    "description": "eZ Publish LegacyStack (4.x)",
+    "description": "eZ Publish LegacyStack (4.x) - changed dependency for eztags to v2",
```
andrerom: No, you shouldn't change this.

andrerom:
The one commit you need here should be done on master, and issue + better description of PR needed.

PR #1214, opened 2015-11-10
Fix EZP-25077: invalid image refs on trashed content with empty images

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-25077

#### Problem: 
due to a bug on ezimage datatype on <= 5.3.2 , trashing a content with an empty image would result in an attribute with the following:
```
<ezimage serial_number="1" is_valid="" filename="c992d69f74dcf51ff04a051ef0d93d6b." suffix="" basename="c992d69f74dcf51ff04a051ef0d93d6b" dirpath="/trashed" url="/trashed/c992d69f74dcf51ff04a051ef0d93d6b." original_filename="" ... 
```

This has been resolved, but not at the DB level.
With the new alias handling mechanism in 5.4.4, operations on these files (such as clearing the trash) will result in a fatal exception.

#### Solution:
Add a script to cleanup these invalid references in empty, trashed, ezimage attributes.

#### Notes:
The script should be able to handle thousands of attributes due to it's simple operation, in any case a limit parameter could be added to the query if really necessary.

### DIFF
--- a/update/common/scripts/5.4/fixtrashedimagereferences.php
+++ b/update/common/scripts/5.4/fixtrashedimagereferences.php
@@ -1,0 +1,103 @@
+<?php
+/**
+ * File containing the ${NAME} class.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ */
+
+require_once 'autoload.php';
+$cli = eZCLI::instance();
+
+$script = eZScript::instance(
+    array(
+        'description' => "Fixes invalid references to image files in trashed content. See issue EZP-25077\n",
+        'use-session' => true,
+        'use-modules' => false,
+        'use-extensions' => true
+    )
+);
+$script->startup();
+
+$options = $script->getOptions(
+    "[dry-run][iteration-sleep:][iteration-limit:]",
+    "",
+    array(
+        'dry-run' => 'Dry run'
+    )
+);
+$optDryRun = (bool)$options['dry-run'];
+$optIterationSleep = (int)$options['iteration-sleep'] ?: 1;
+$optIterationLimit = (int)$options['iteration-limit'] ?: 100;
+
+$limit = array(
+    "offset" => 0,
+    "limit" => $optIterationLimit,
+);
+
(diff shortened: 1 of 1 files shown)


yannickroger:
+1
ping @bdunogier @andrerom

bdunogier:
Still a bit unsure about a query without a LIMIT _with a LIKE_ on a string that ain't indexed...

But +1 once that one is clarified/approved by others.

bdunogier:
After talking about it during this friday's meeting, we concluded that there was no way this query could be nice with the DB server.

We think it'd be better to iterate over all ezimage attributes, by batch of X (100 ?), and check the XML in PHP. Then if a record needs to be fixed, issue the update.

This will take more resources from the php server, but much less from the db server.

joaoinacio:
Updated.

The query now does a join with objects in trash, rows are fetched using batch (as other scripts, limit 100 sleep 1 by default), and xml match is done in PHP.

The matching is done with `stripos()` for performance, as opposed to always loading the xml (possibly a bit cleaner).

ping @bdunogier

yannickroger:
still +1 ping @andrerom @bdunogier

update/common/scripts/5.4/fixtrashedimagereferences.php
```diff
@@ -0,0 +1,103 @@
...
+
+if ( $optDryRun ) {
+    $cli->warning( "dry-run mode" );
+}
+
+/**
+ * Fix xml attributes for empty images that have been trashed.
+ *
+ * @param $imageAttribute
+ * @param $optDryRun bool   if in dry-run mode (do not update BD row)
```
glye: DB row :)

glye:
+1 apart from comment nitpick

bdunogier:
I wouldn't really have bothered with the `stripos` vs xml: what mattered is/was the load on the database. This script could be executed from a completely independent instance, and it would still work.

But it's fine :-)

update/common/scripts/5.4/fixtrashedimagereferences.php
joaoinacio: good catch, thanks!

joaoinacio:
Done. thanks for the review!

PR #1215, opened 2015-11-18
Fix EZP-24710: Paragraphs are stripped on the front-end if an XML blo…

glye:
…ck is published with the editor disabled

> Fixes https://jira.ez.no/browse/EZP-24710
> Status: Ready for review & ridicule

Ugly inefficient solution with hardcoded ezoe dependency, should not be merged as is, but seems to work fine. PR to make sure all tests pass.

### DIFF
--- a/kernel/classes/datatypes/ezxmltext/handlers/input/ezsimplifiedxmlinput.php
+++ b/kernel/classes/datatypes/ezxmltext/handlers/input/ezsimplifiedxmlinput.php
@@ -105,6 +105,13 @@ class eZSimplifiedXMLInput extends eZXMLInputHandler
             $contentObject->appendInputRelationList( $parser->getRelatedObjectIDArray(), eZContentObject::RELATION_EMBED );
             $contentObject->appendInputRelationList( $parser->getLinkedObjectIDArray(), eZContentObject::RELATION_LINK );
 
+            // Fixes EZP-24710 by processing the XML through ezoe
+            $tempOutputHandler = new eZXHTMLXMLOutput( $xmlString, false, $contentObjectAttribute );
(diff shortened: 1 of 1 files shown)


yannickroger:
Just in case you want to run the test locally, we have a prehistoric doc covering it http://www.ezpedia.org/ez/testing_ez_publish_test_system

glye:
@yannickroger Yes, I know, just wanted to make sure all applicable tests are run (turns out there were not that many, given how fast it was).

andrerom:
@dougplant unknown atm, ezxmltext handlers are one of the more complex parts of the system, and we need to solve it in the handler itself and not use oe handler for this.

kernel/classes/datatypes/ezxmltext/handlers/input/ezsimplifiedxmlinput.php
```diff
@@ -105,6 +105,13 @@ function validateInput( $http, $base, $contentObjectAttribute )
             $contentObject->appendInputRelationList( $parser->getRelatedObjectIDArray(), eZContentObject::RELATION_EMBED );
             $contentObject->appendInputRelationList( $parser->getLinkedObjectIDArray(), eZContentObject::RELATION_LINK );
 
+            // Fixes EZP-24710 by processing the XML through ezoe
+            $tempOutputHandler = new eZXHTMLXMLOutput( $xmlString, false, contentObjectAttribute );
```
joaoinacio: should `contentObjectAttribute` be a var?

glye: Nope. Fixing...

dougplant:
Ok, thanks for that.

PR #1216, opened 2015-11-20
Fix EZP-25109: purge stale binary files when removing from cluster

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-25109

When updating the first draft of a content the binary files (videos/etc) are removed but not purged from the cluster.
This is an alternative to having to run the dfscleanup script.

### DIFF
--- a/kernel/classes/datatypes/ezbinaryfile/ezbinaryfiletype.php
+++ b/kernel/classes/datatypes/ezbinaryfile/ezbinaryfiletype.php
@@ -185,7 +185,10 @@ class eZBinaryFileType extends eZDataType
                 $file = eZClusterFileHandler::instance( $filePath );
 
                 if ( $file->exists() and count( $binaryObjectsWithSameFileName ) < 1 )
+                {
                     $file->delete();
+                    $file->purge();
+                }
             }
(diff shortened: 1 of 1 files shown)


yannickroger:
How does it affect performances ?

joaoinacio:
@yannickroger : 
Good question, in theory this operation shouldn't be as common as others so it shouldn't cause a significant performance degradation in most operations;
However, the nature of the binaryfiles (which could be large videos, for example) can cause some impact (for example, when removing archived versions above the max. threshold).

Other than that, I don't recall any situations where this would be called multiple times on the same request (unless, of course, for contents with multiple binary files).

Any additional feedback is welcome

yannickroger:
Because my main concern is that this could actually be a feature and not a bug. Files being actually deleted by a cronjob to maximize performances as stated in this documentation: https://doc.ez.no/eZ-Publish/Technical-manual/5.x/Features/Clustering/Maintenance

joaoinacio:
Update: The clusterpurge maintenance script removes the files when used with the appropriate scope (binaryfile), so closing this PR as discussed

PR #1217, opened 2015-11-24
Fix EZP-25065: non-translatable object relation not updated.

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-25065

#### Problem:
When editing content with multiple languages and a non-translatable object relation, updating the relation object will cause the object to incorrectly display both the previous and the new relation.

#### Solution:
When the relation attribute is non-translatable, remove any existing relation when storing a new one, regardless of the number of current translations.

### DIFF
--- a/kernel/classes/datatypes/ezobjectrelation/ezobjectrelationtype.php
+++ b/kernel/classes/datatypes/ezobjectrelation/ezobjectrelationtype.php
@@ -158,23 +158,34 @@ class eZObjectRelationType extends eZDataType
         $contentObjectID = $contentObjectAttribute->ContentObjectID;
         $contentObjectVersion = $contentObjectAttribute->Version;
 
-        $obj = $contentObjectAttribute->object();
-        //get eZContentObjectVersion
-        $currVerobj = $obj->version( $contentObjectVersion );
-        // get array of language codes
-        $transList = $currVerobj->translations( false );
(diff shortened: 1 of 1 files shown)


jjCavalleri:
This fix worked in my tests

kernel/classes/datatypes/ezobjectrelation/ezobjectrelationtype.php
```diff
@@ -158,23 +158,33 @@ function storeObjectAttribute( $contentObjectAttribute )
...
-        $currVerobj = $obj->version( $contentObjectVersion );
-        // get array of language codes
-        $transList = $currVerobj->translations( false );
-        $countTsl = count( $transList );
+        $contentObject = $contentObjectAttribute->object();
+
+        // check if previous relation(s) can be removed according to existing translations
+        if ( $contentObjectAttribute->contentClassAttributeCanTranslate() )
+        {
+            // get eZContentObjectVersion
```
yannickroger: Nitpick: you can typehint the variable in the comment with eZContentObjectVersion it will provide the same info and make it IDE friendly for completion.

yannickroger:
+1

joaoinacio:
Updated, thanks!.
ping @andrerom @bdunogier

glye:
+0.5

andrerom:
+0.5

PR #1218, opened 2015-11-24
Fix EZP-24783: only clear cache if needed in sort operations

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-24783

##### Problem:
Some node sort operations (either through content/action, content operation or ezjscore) clear cache without checking if it is enabled or not.


andrerom:
+1

yannickroger:
+1

PR #1219, opened 2015-11-25
EZP-25077: fix incorrect order of trashed image validity check.

joaoinacio:
Somehow the string comparison for matching invalid/empty, trashed images was incorrect, this fixes it.


yannickroger:
Merge with no reviews ? is it an accident ? @joaoinacio

joaoinacio:
@yannickroger: This fixes an unfortunate error in https://github.com/ezsystems/ezpublish-legacy/pull/1214 , which has been otherwise reviewed. 
I did not believe the change here needed any further reviewing besides the above, as there is no additional logic or risk involved.

yannickroger:
no problem :)

PR #1220, opened 2015-11-26
Fix EZP-25156: clusterpurge requires a valid list of scope(s)

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-25156

If no `--scopes=` parameter is provided, exit with a warning instead of proceeding and failing silently with a SQL error ( `IN()` operator with an empty string ).

### DIFF
--- a/kernel/private/classes/ezscriptclusterpurge.php
+++ b/kernel/private/classes/ezscriptclusterpurge.php
@@ -65,6 +65,12 @@ class eZScriptClusterPurge
     {
         $cli = eZCLI::instance();
 
+        if ( empty( $this->optScopes ) )
+        {
+            $cli->warning( "No scope(s) to purge provided, exiting..." );
+            return;
+        }
+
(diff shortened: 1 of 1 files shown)


kernel/private/classes/ezscriptclusterpurge.php
```diff
@@ -65,6 +65,12 @@ public function run()
     {
         $cli = eZCLI::instance();
 
+        if ( empty( $this->optScopes ) )
```
gggeek: I kind of expected all copes to be purged?

joaoinacio: @gggeek : indeed, but if none is defined the default `false` is used.
So this PR just checks it as it doesn't intend any change in behavior (another possibility would be to use all)

yannickroger:
+1
I agree with @gggeek it would have been better that no scopes means to clear all. But to me BC is more important.

andrerom:
+1

PR #1221, opened 2015-11-30
Fix EZP-18202: Impossible to clear previously populated ISBN field

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-18202

#### Problem:
When clearing an isbn13 field that was already filled in (and the field is not required), saving will not actually clear the field / save the empty value.

This makes it so the empty value is actually saved.


crevillo:
Finally :). +1.

glye:
+1

yannickroger:
+1

PR #1222, opened 2015-12-01
Fix EZP-25058: md5 string comparision with "==" fails, if the strings…

yannickroger:
… begin with "0e"

Link: https://jira.ez.no/browse/EZP-25058

## Description
With `==` some md5 strings are considered as float in the string comparison, see detailed info in the Jira.

## Test
Basic sanity on users.

### DIFF
--- a/kernel/classes/datatypes/ezuser/ezuser.php
+++ b/kernel/classes/datatypes/ezuser/ezuser.php
@@ -264,7 +264,7 @@ class eZUser extends eZPersistentObject
         $this->setAttribute( "email", $email );
         $this->setAttribute( "login", $login );
         if ( eZUser::validatePassword( $password ) and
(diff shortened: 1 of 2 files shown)


yannickroger:
Ready for reviews

andrerom:
+1

yannickroger:
I did, I also checked for easy ones to find (usage of md5 within the same file), but when variables are passed to function, it's very tricky and time consuming to investigate all of that. I has been working fine for a long time now, so I only fixed this edge case.

andrerom:
ok

glye:
+1

yannickroger:
Added a missing part of the fix in https://github.com/ezsystems/ezpublish-legacy/commit/3191e3dccbd2cd3f350037402f024abd2950ca6b

PR #1224, opened 2015-12-16
Fix EZP-25927 ezpgenerateautloads.php expects vendor folder in a place where is not

crevillo:
Fixes https://jira.ez.no/browse/EZP-25297

vendor and app/ezpublish folder are in the same level, so i think actually legacyVendorDir definition was wrong. 

ping @emodric @bdunogier @andrerom

### DIFF
--- a/bin/php/ezpgenerateautoloads.php
+++ b/bin/php/ezpgenerateautoloads.php
@@ -18,7 +18,7 @@ if ( file_exists( "config.php" ) )
 //{
 $appName = defined( 'EZP_APP_FOLDER_NAME' ) ? EZP_APP_FOLDER_NAME : 'ezpublish';
 $appFolder = getcwd() . "/../$appName";
-$legacyVendorDir = getcwd() . "/vendor";
+$legacyVendorDir = getcwd() . "/../vendor";
 
 $baseEnabled = true;
 // Bundled


emodric:
I don't think this is right.

`getcwd() . "/vendor"` is supposed to be used when legacy is running in pure legacy context (that is, when webroot is legacy folder).

crevillo:
well, yes, but that is how it's right now in master... i'm not changing that. i've seen your pull request where you are using ```__DIR__``` instead, so should i probably wait to have that merged?

emodric:
Well yes, but how it's right now in the master is the correct way, I think.

About `__DIR__`, I can change it to use `getcwd()`, it doesn't really matter, right?

crevillo:
please see my edit. thing is it fails as it's now... when running the script with app/console... 
so i don't think is completely right... maybe should we do checks there?

emodric:
I tnink it fails because the patch is not needed.

`$legacyVendorDir` is used when running in **pure** legacy context (when current working directory is eZ Publish legacy dir). In that case, when running `composer install` to install Zeta Components, vendor folder will be in `vendor` subdir of eZ Publish legacy, thus `getcwd() . "/vendor"` is right.

@lolautruche will probably know more.

crevillo:
@emodric now i see what you mean. ok. got it. but then, something needs to be done on installed, because if not, all those if else leads you to https://github.com/ezsystems/ezpublish-legacy/blob/master/bin/php/ezpgenerateautoloads.php#L52 and there's the crash. 
with my patch you enter https://github.com/ezsystems/ezpublish-legacy/blob/master/bin/php/ezpgenerateautoloads.php#L44 and legacy autolad gets regenerated.

emodric:
Did you run `composer install` before generating autoloads?

Current master works fine for me. See here: https://gist.github.com/emodric/51504521bba888c9e012

crevillo:
Nope. This happened when running the script after ```composer require ezsystems/legacy-bridge```.  That is, adding legacy to a ezplatform install.

emodric:
So you're not running in legacy context? Then this patch is invalid. Your error will be solved with #1223.

crevillo:
ah, ok then.

PR #1226, opened 2015-12-17
Fix for Issue EZP-25301

pkamps:
Link: https://jira.ez.no/browse/EZP-25301

With eZContentUpload you can use custom UploadHandler. eZContentUpload will then call the method 'handleFile' of your custom UploadHandler. One of the method parameters is the original file name "originalFilename". eZContentUpload is not sending the correct value for this parameter -- this PR is fixing it.

### DIFF
--- a/kernel/classes/ezcontentupload.php
+++ b/kernel/classes/ezcontentupload.php
@@ -246,7 +246,7 @@ class eZContentUpload
         // The handler will be responsible for the rest of the execution.
         if ( is_object( $handler ) )
         {
-            $originalFilename = $nameString ? $nameString : $filePath;
+            $originalFilename = $nameString != '' ? $nameString : $filePath;
             return $handler->handleFile( $this, $result,
                                          $filePath, $originalFilename, $mimeData,
                                          $location, $existingNode );
--- a/kernel/classes/ezcontentupload.php
+++ b/kernel/classes/ezcontentupload.php
@@ -246,7 +246,7 @@ class eZContentUpload
         // The handler will be responsible for the rest of the execution.
         if ( is_object( $handler ) )
         {
-            $originalFilename = $filePath;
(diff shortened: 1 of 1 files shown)


kernel/classes/ezcontentupload.php
```diff
@@ -246,7 +246,7 @@ function handleLocalFile( &$result, $filePath, $location, $existingNode, $nameSt
         // The handler will be responsible for the rest of the execution.
         if ( is_object( $handler ) )
         {
-            $originalFilename = $filePath;
+            $originalFilename = $nameString ? $nameString : $filePath;
```
crevillo: shouldn't we check if $nameString is ```!= ''```?

yannickroger:
Hi,
If you want to get this merged, you should:
-  add some steps to reproduce in your jira issue.
- fill up the description of the PR with what it is about and how you tested it.
- squash your commits.

kernel/classes/ezcontentupload.php
pkamps: yes, that's true

pkamps:
Steps added to the jira issue.
Description to the PR added.

Could you explain why how and why I would need to squash my commits, thanks?

yannickroger:
Regarding why squash commit, it is to keep a clean and readable git history.
So, your fix should be in a commit called something like: `Fix EZP-25031: eZContentUpload is not providing original file name to custom UploadHandler`

How to clean it up, I can help you here or you can ping me on slack. (If you prefer the RTFM way :) what to look for is `git rebase -i` and `push --force`)

pkamps:
Closed in favor of https://github.com/ezsystems/ezpublish-legacy/pull/1228 which only contains a single commit.

Thanks for explaining the squashing. The changes are so simple that I decide to create a new pull request.

PR #1229, opened 2015-12-23
Fix EZP-25331: Specify table name for path attribute filter

peterkeung:
https://jira.ez.no/browse/EZP-25331

Ensures that the field name does not conflict with other tables

### DIFF
--- a/kernel/classes/ezcontentobjecttreenode.php
+++ b/kernel/classes/ezcontentobjecttreenode.php
@@ -1022,7 +1022,7 @@ class eZContentObjectTreeNode extends eZPersistentObject
                     {
                         case 'path':
                         {
-                            $filterField = 'path_string';
+                            $filterField = 'ezcontentobject_tree.path_string';
                             $filterFieldType = 'string';
                         } break;
                         case 'published':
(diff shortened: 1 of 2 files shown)


emodric:
@peterkeung How exactly this breaks with eZ Tags? As far as I can see, eZ Tags uses table aliases (i1, i2, i3), to prevent exactly this from happening.

peterkeung:
Table aliases don't prevent a clash on the same column name if the table (alias or not) is not specified. I don't think it's the fault of eZ Tags, though, and it should be prevented at the eZ kernel level.

andrerom:
+1

peterkeung:
Ping @benkmugo @alconasoftware @dougplant @pkamps @thiagocampoviana @ernestob for review

ernestob:
***
:+1:  It looks good
***

pkamps:
You're making the DB field name more specific by prepending the 'ezcontentobject_tree'. I cannot see how that is generating a wrong or broken SQL statement. We can be certain that this DB table is always part of the SQL query, right? 

In the same function it does exactly the same for other filter keys:

case 'node_id':
{
    $filterField = 'ezcontentobject_tree.node_id';
} break;

Again, I cannot see how it would break fetches.

+1

crevillo:
+1. this could prevent crashes if you add your own extended filter with a table containing a path_string. it may say path_string is ambigous...

andrerom:
> You're making the DB field name more specific by prepending the 'ezcontentobject_tree'. I cannot see how that is generating a wrong or broken SQL statement. We can be certain that this DB table is always part of the SQL query, right?

In other generic fetch helper functions there is usually a argument to specify object and node table, so in those cases it could. But I also saw it is already hardcoding it several places below so I'm fully +1

guillaumelecerf:
+1
BTW you could probably do the same for `modified_subnode` and `depth` (see the unit test)

guillaumelecerf:
https://github.com/ezsystems/ezpublish-legacy/commit/2d6a887dff41beddc15578d152505901dbcb2eb3

peterkeung:
+1 to that more complete version

guillaumelecerf:
See https://github.com/ezsystems/ezpublish-legacy/pull/1238

yannickroger:
Closed in favor of #1238

PR #1230, opened 2015-12-24
FIx wrong path creation in css pack

Alexnder:
If you have several files of one font, for ex.: `url(fonts/font.woff)` `url(fonts/font.woff2)`
It's pack to this:
```
url(/design/siteaccess/stylesheets/fonts/font.woff)
url(/design/siteaccess/stylesheets//design/siteaccess/stylesheets/fonts/font.woff2)
```

The commit fix this bug.


andrerom:
+0,5 as we don't have any unit tests here so a full +1 will need a bit more testing from community *(one alternative is that we merge and await backports until the patch has been in circulation for some months to await possible feedback)*

PR #1231, opened 2016-01-05
Fix EZP-25065: non-translatable object relation list not updated.

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-25065

Follow up on #1217


andrerom:
+1

yannickroger:
+1

PR #1227, opened 2015-12-17
Fix EZP-25292: fix stale relations for non-translatable attributes.

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-25292

PR #1232, opened 2016-01-06
Fix EZP-25292: additional check for relationlist xml

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-25292

This adds an additional check for non-translatable relation(s) cleanup script in #1227

For cases where the `data_text` column of relationlist attribute is `null`, for whatever reason, the script would try to access a non-object when processing the relations xml.

Ref: `eZObjectRelationListType` @ https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/classes/datatypes/ezobjectrelationlist/ezobjectrelationlisttype.php#L1415

### DIFF
--- a/update/common/scripts/5.4/cleanuntranslatablerelations.php
+++ b/update/common/scripts/5.4/cleanuntranslatablerelations.php
@@ -53,8 +53,15 @@ if ( $optDryRun ) {
  */
 function parseRelationListIds( $relationListXml )
 {
+    if ( empty( trim( $relationListXml ) ) )
+        return array();
+
     $xml = simplexml_load_string( $relationListXml );
     $list = $xml->{'relation-list'};
+
+    if ( !( $list instanceof SimpleXMLElement ) ) {
+        return array();
+    }
 
     $contentIds = array();
     foreach ( $list->children() as $item ) {


update/common/scripts/5.4/cleanuntranslatablerelations.php
```diff
@@ -53,12 +53,17 @@
  */
 function parseRelationListIds( $relationListXml )
 {
+    if ( empty( trim( $relationListXml ) ) )
+        return;
```
andrerom: function seems to always have returned array before this, so please stick to that.

joaoinacio: Yes, it should return array for consistency. (the `empty()` check on the function result would still skip it anyway)
updating, thanks!

update/common/scripts/5.4/cleanuntranslatablerelations.php
```diff
@@ -53,12 +53,17 @@
...
+    if ( empty( trim( $relationListXml ) ) )
+        return array();
+
     $xml = simplexml_load_string( $relationListXml );
     $list = $xml->{'relation-list'};
 
     $contentIds = array();
-    foreach ( $list->children() as $item ) {
-        $contentIds[] = (int)$item['contentobject-id'];
+    if ( $list ) {
```
bdunogier:
```
if ( !$list ) {
    return array();
}
```

Also, try to use a strict test, against the "empty" value we might get from `$xml->{'relation-list'}`.

yannickroger:
+1 with @bdunogier 's comment taken into account.

update/common/scripts/5.4/cleanuntranslatablerelations.php
```diff
@@ -53,9 +53,16 @@
  */
 function parseRelationListIds( $relationListXml )
 {
+    if ( empty( trim( $relationListXml ) ) )
+        return array();
+
     $xml = simplexml_load_string( $relationListXml );
     $list = $xml->{'relation-list'};
 
+    if ( ! $list instanceof SimpleXMLElement ) {
```
bdunogier: CS (no space between `!` and `$list`.

yannickroger: Looks valid to me

joaoinacio: Good point, I find negating the `instanceof` operator strange imho ... 
It should read `( ! ( $var instanceof Class ) )`, but the instanceof has higher precedence hence no parenthesis needed; I just tend to read `!$var instanceof Class` the wrong way (negating first)...

As for legacy CS I don't know the exact rule atm, but a search shows both these forms:
```
if( !$user instanceof eZUser )
if ( !( $user instanceof eZUser ) )
```

So... I'll just remove the space.

gggeek: lol, using '! $x instanceof $y' is one of the things I always do and cross my fingers hoping it will work :-D so parenthesis are fine with me.
If not using parenthesis, please at least never use '!$x instaceof $y' and keep the initial space, as otherwise brain goes shortcircuit

bdunogier:
+1, though

guillaumelecerf:
`PHP Fatal error:  Can't use function return value in write context in /home/user/ezpublish/update/common/scripts/5.4/cleanuntranslatablerelations.php on line 56`

PHP 5.4.45

See PR #1234

PR #1233, opened 2016-01-11
EZP-27804: Constructor refactoring (PHP4 to PHP5 __construct())

jeromegamez:
https://jira.ez.no/browse/EZP-27804

This PR supersedes PR #789, here is the original (and still fitting) PR description:

---

Due to eZ Publish's history, most of the classes use PHP4 style constructors in the form

``` php
class MyClass
{
    function MyClass()
    {
        // Constructor
        $this->MyParentClass();
        // More code
    }
}
```

instead of

``` php
class MyClass
{
    public function __construct()
    {
        // Constructor
        parent::__construct();
        // More code
    }
}
```

In our modern times, this is problematic due to the following reasons:
- If you want to use the parent's constructor, you have to call it explicitely
- Derived classes call the parent constructor by its name instead of `parent::__construct()` which makes the refactoring of class names cumbersome
- We don't know if and when a future version of PHP will stop supporting the PHP4 constructor methods

This pull request changes this by doing the following:
- Replace every PHP4 style constructor with the PHP5 style `public function __construct()`
- Replace every call to the parent's constructor with `parent::__construct()`
- Delete all empty constructors where it makes sense (e.g. when the parent has no constructor or overriding the parent's constructor wouldn't make sense)
- Delete all constructors that just call the parent constructor
- For the more "popular" classes that are most likely to be used in 3rd party extensions, the new constructor is just added, and the old style constructor is marked as deprecated (although I personally would prefer not to do that, 3rd party extensions shouldn't be too old and outdated)
- Replace every other direct call of the old style constructor

If you think more classes should keep the old style constructor (marked as deprecated) in addition to the new one, please let me know.

Cheers
:octocat: Jérôme


jeromegamez:
@patrickallaert @lolautruche @gggeek @pkamps 

In light of PHP7 being released and @patrickallaert's https://github.com/patrickallaert/ezpublish-legacy-php7, would you perhaps be open to reconsider this PR? I just rebased it with the current master.

Of course, BC is still an issue, but by now, this wouldn't be a [premature optimization](#issuecomment-26416819) any more, but required to be able to run the old stack on PHP 7.

I also cherry picked @patrickallaert's [exception handler change](https://github.com/patrickallaert/ezpublish-legacy-php7/commit/ba26b67a9afce5395cec7266fb97ee86b0e60300).

And sorry for the repeated ping.

jeromegamez:
As a reminder: the PHP4-Style constructors are still present and just call `self::construct(…)`, so backwards compatibility should not be an issue.

See an example here: https://3v4l.org/f5eOT

gggeek:
+1 FWIW (I did not go over all of the files with a fine comb to check for typos, did you use a script for the changes or do them all by hand?)

Another unrelated PR I have been thinking about for a while: use everywhere self::dostuff() instead of currentClassName::dostuff() (that, and change all doxygen comments to phpdoc comments). Would there be consensus on those?

lib/ezutils/classes/ezexecution.php
```diff
@@ -177,7 +177,7 @@ static public function registerShutdownHandler( $documentRoot = false )
      * @params Exception the exception
      * @return void
      */
-    static public function defaultExceptionHandler( Exception $e )
+    static public function defaultExceptionHandler( Throwable $e )
```
emodric: This will break on PHP 5 since it doesn't have `Throwable` class. Typehint should probably be removed altogether.

jeromegamez:
Does this mean this PR is alive again?

emodric:
> Does this mean this PR is alive again?

I don't think so :) I'm not affiliated with eZ Systems despite "eZ Systems member" badge :)

andrerom:
> Does this mean this PR is alive again?

No not really, but we can close this and reopen a new one with 16c1f829f24c14a54bd84f2d8854513df1b7efc9, but like @emodric implied removing the type hint and check type inside the function in a way that works across versions.
With that and some other commits laying around we can get it to run on php7 *(at own risk, extensions and all)* with deprecated warnings disabled.

pkamps:
Do you plan to submit an updated version? I was about to merge this pull request into mugo fork.

jeromegamez:
Oops, I just thought the PR was dead and it had merge conflicts anyway - but I just restored the previous state so that you can use it again. Sorry!

jeromegamez:
@pkamps I also just resolved the conflicts (the Github Interface is nice :))

pkamps:
Hi,

* about "keeping the old style constructor", could you explain (give an example) when that's required to have
* in a few child classes of eZPersistentObject the constructor had the $row parameter optional (for example https://github.com/ezsystems/ezpublish-legacy/pull/1233/files#diff-3934f1e762d1519c3b073865eb86cc56). eZPersistentObject requires the parameter. So there was some inconsistency in the code before. Removing the child constructor is fixing it. But we could end up with a PHP warning, right? Some code might create a child object instance without the $row parameter. But I don't think that it is a big deal.
* https://github.com/ezsystems/ezpublish-legacy/pull/1233/files#diff-b22ceed14b61605e9442c5e89d418448, you're removing the constructor. The parent constructor is not empty. Maybe we should keep an empty constructor in eZTextFileUser?
* Same issue on https://github.com/ezsystems/ezpublish-legacy/pull/1233/files#diff-4574877c8e10b80bd368c40c06ba2606 - parent constructor is not empty
* https://github.com/ezsystems/ezpublish-legacy/pull/1233/files#diff-87b099d6cf41f15f472926dc372a0c2d, it still has the name constructor - is that an overlook? 
* https://github.com/ezsystems/ezpublish-legacy/pull/1233/files#diff-7ad0d6b2ce913e376cf1322ffdf3faba - looks like some utf8 chars broke

Thanks for this contribution (this pull request).

andrerom:
@jeromegamez / @pkamps Now that we do legacy releases again this can make sense as long as it does not break to many extensions, anyone up for finishing this PR? Could be part of 2017.10 release ;)  /cc @emodric

jeromegamez:
I will give it a new, fresh shot next week from the current master and keep the issues in mind :+1:

jeromegamez:
> @pkamps: about "keeping the old style constructor", could you explain (give an example) when that's required to have

If I remember correctly, the intention was to not break 3rd party extensions that call the old style constructor directly. As some time has passed now, I think it should be safe to remove the old style constructors completely.

emodric:
> If I remember correctly, the intention was to not break 3rd party extensions that call the old style constructor directly. As some time has passed now, I think it should be safe to remove the old style constructors completely.

I wouldn't say so. In time passed from previous attempt at this PR, I would say 99% of legacy extensions were never touched so chance that they removed calls to legacy style constructors is pretty low.

I've searched a bit, and for example, every datatype extension out there I could find includes a call to `$this->eZDataType(...)` it it's constructor.

jeromegamez:
@emodric In that case, I'll keep everything as it is :)
@pkamps I fixed the issues you pointed out, thank you for looking through everything!

@emodric @pkamps @andrerom If everything looks okay, I would like to rebase the current master onto the PR branch to keep things clean and in order, if it's okay for you.

emodric:
> @emodric @pkamps @andrerom If everything looks okay, I would like to rebase the current master onto the PR branch to keep things clean and in order, if it's okay for you.

Wouldn't want it any other way ;)

pkamps:
I reviewed your commits. Looks good now.

jeromegamez:
@andrerom @emodric @pkamps Rebase complete, this should now be ready to go.

andrerom:
@jeromegamez any changes in the rebase that we should be aware of? :)

jeromegamez:
@andrerom None that I am aware of. I had some conflicts during the rebase, but they were easy to resolve (one was a PHPDoc, and the others were about parameters being passed by reference that shouldn't). I looked over everything again in the "Files Changed" tab and didn't notice anomalies, and the Test Suite passed, so I am optimistic that everything should work out (fingers crossed)

andrerom:
@jeromegamez Please be aware you might need to help us add some more bc functions once this is merged and strange things pop up on Slack :)

emodric:
@andrerom Looks good to me, but definitely with counting on your last comment about further work :)

andrerom:
ok with you @jeromegamez? ;)

jeromegamez:
Absolutely okay with that!

andrerom:
Cool, one last thing before we merge, could you create an jira issue for this?

Would for instance the following be correct to say? `Avoid PHP7 deprecation errors when running legacy` or `PHP7 logs deprecation errors when running legacy`

jeromegamez:
Here it is: https://jira.ez.no/browse/EZP-27804 - I updated the PR title and description accordingly

andrerom:
As we are some 3 weeks aways from 2017.10 release, lets keeps our ears open for issues this might be causing for those using master. Thanks so far @jeromegamez !

pkamps:
Celebration!!

emodric:
@andrerom @blankse I think I bumped into the first issue with this:

`Too few arguments to function eZPersistentObject::__construct(), 0 passed in /var/www/html/netgensite/ezpublish_legacy/kernel/content/versionview.php on line 191 and exactly 1 expected in /var/www/html/netgensite/ezpublish_legacy/kernel/classes/ezpersistentobject.php on line 52`

Line 191 is a simple `new eZContentObjectTreeNode()`

thiagocamposviana:
When trying to install the packages I needed to change ezsrRatingObject constructor to public (in var/storage/packages) and ezdemoInstaller:

new eZSiteInstaller( $parameters );

to

parent::__construct( $parameters );

andrerom:
> @andrerom @blankse I think I bumped into the first issue with this:

I guess the ping was for @jeromegamez :)


Any of you see how we can add compat to avoid this and https://github.com/ezsystems/ezfind/pull/212 ?

emodric:
> I guess the ping was for @jeromegamez :)

Yes, sorry about that :)

> Any of you see how we can add compat to avoid this and ezsystems/ezfind#212 ?

Hm... Not without leaving the old style constructor in. It wasn't removed for critical classes like `eZDataType`, but that should be the case for a limited set of classes that are official extension points.

I don't see a problem with adapting other code on need by need basis.

jeromegamez:
I think #1330 is an issue that should be definitely addressed/merged, for https://github.com/ezsystems/ezfind/pull/212 and similar issues I agree with @emodric: not without restoring the old style constructors/reverting the changes from this PR. Adapting the code in extensions would be the way that I personally would prefer, though

PR #1234, opened 2016-01-25
Fix EZP-25292: fix additional check

guillaumelecerf:


andrerom:
+1, ping @joaoinacio

update/common/scripts/5.4/cleanuntranslatablerelations.php
```diff
@@ -53,7 +53,8 @@
  */
 function parseRelationListIds( $relationListXml )
 {
-    if ( empty( trim( $relationListXml ) ) )
+    $trimmedRelationListXml = trim( $relationListXml );
```
guillaumelecerf: Maybe should be `$relationListXml = trim( $relationListXml );` ?

joaoinacio:
It's OK, the alternative option would be to leave master as-is (php 5.5, cleaner IMHO) and for 5.4 perform the comparison using `if ( trim( $relationListXml ) == false )` as PHP officially recommends for this case.

guillaumelecerf:
@joaoinacio : what would you prefer ?

andrerom:
from my side: lets try to avoid having differences between master and backports, also note that master for legacy supports php 5.3.3 and up as it won't receive further major releases.

update/common/scripts/5.4/cleanuntranslatablerelations.php
andrerom: can do that

joaoinacio:
OK given the above it's a no-brainer, +1 as-is

guillaumelecerf:
Thanks.

PR #1236, opened 2016-02-08
 Fix EZP-25455: eZOracle: Link management doesn't show objects

fwoldt:
Link: https://jira.ez.no/browse/EZP-25455

Oracle uses short_names in ezPersistentObject to access DB field values.
The Bugfix ensures that the correct replacement of the attribute short_names with the longnames will be used
for objectcreation!
The first &$row is used for correct replacement and the second $row fixes an issue where rsults are correctly used for objeccreation
(otherwise e.g. worflow/grouplist was not shown correctly  3 groups but only 2 where displaxed one was overriden)

### DIFF
--- a/kernel/classes/ezpersistentobject.php
+++ b/kernel/classes/ezpersistentobject.php
@@ -951,7 +951,7 @@ class eZPersistentObject
         if ( is_array( $rows ) )
         {
             $db = eZDB::instance();
-            foreach( $rows as $row )
+            foreach( $rows as &$row )
             {
                 self::replaceFieldsWithLongNames( $db, $objectDefinition['fields'], $row );
             }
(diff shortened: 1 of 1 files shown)


joaoinacio:
+1, would it just be possible to shorten the commit message (removing 'ezpublish_legacy' from title)?
Thanks!

yannickroger:
+1
@bdunogier can you also check it out ?

kernel/classes/ezpersistentobject.php
```diff
@@ -951,7 +951,7 @@ public static function handleRows( $rows, $class_name, $asObject )
         if ( is_array( $rows ) )
         {
             $db = eZDB::instance();
-            foreach( $rows as $row )
+            foreach( $rows as &$row )
```
bdunogier: Do wee need to loop at all ? `replaceFieldsWithLongNames()` will already iterate on the fields passed as the 3rd argument. I'm thinking that replacing the whole `foreach` with `self::replaceFieldsWithLongNames( $db, $objectDefinition['fields'], $rows )` would just do the same thing.

bdunogier:
+1

yannickroger:
Thanks @fwoldt

fwoldt:
Thank you for merging :-)

PR #1237, opened 2016-02-09
FIX EZP-25456 - ezp_legacy: eZPersistentObject $this->$def["functions"][$attr]();

fwoldt:
Bugfix in eZPersistentObject changing
$this->$def["functions"][$attr]();
 => 
$def["functions"][$attr]();


andrerom:
This seems to break fix done in #475.

PR #1240, opened 2016-03-03
Refresh eZExpiryHandler instance after change siteaccess.

Opencontent:
This pull requeste is needed in multisite-multidb installations. The expiry values are stored in each var dir, rather than globally where the values of the various db may conflict.

### DIFF
--- a/lib/ezutils/classes/ezexpiryhandler.php
+++ b/lib/ezutils/classes/ezexpiryhandler.php
@@ -160,6 +160,15 @@ class eZExpiryHandler
 
         return $GLOBALS['eZExpiryHandlerInstance'];
     }
+    
+    static function refresh()
+    {
+        if ( isset( $GLOBALS['eZExpiryHandlerInstance'] ) )
+        {
+            unset( $GLOBALS['eZExpiryHandlerInstance'] );
+        }
+        return self::instance();
+    }
 
     /**
      * Checks if a shared instance of eZExpiryHandler exists
--- a/kernel/classes/ezsiteaccess.php
+++ b/kernel/classes/ezsiteaccess.php
@@ -552,6 +552,8 @@ class eZSiteAccess
             eZUpdateDebugSettings();
             eZDebugSetting::writeDebug( 'kernel-siteaccess', "Updated settings to use siteaccess '$name'", __METHOD__ );
         }
+        
+        eZExpiryHandler::refresh();
 
         return $access;
     }


lib/ezutils/classes/ezexpiryhandler.php
```diff
@@ -160,6 +160,15 @@ static function instance()
...
         return $GLOBALS['eZExpiryHandlerInstance'];
     }
+    
+    static function refresh()
+    {
+        if ( isset( $GLOBALS['eZExpiryHandlerInstance'] ) )
+        {
+            unset( $GLOBALS['eZExpiryHandlerInstance'] );
+        }
+        return self::instance();
```
bdunogier: Why does it return an instance ?

bdunogier:
> With this patch the expiry values are stored in each var dir rather than globally.

Not sure I get where this change is done :-)

lib/ezutils/classes/ezexpiryhandler.php
Opencontent: Isn't usefull?

Opencontent:
@bdunogier  yes my comment is a bit cryptic :)
You need to properly configure the site.ini FileSettings VarDir (it is assumed that each "group" of site access shares the same var dir). 
If your site.ini/FileSettings/CacheDir settings is relative, the expiry.php file will be stored in VarDir/CacheDir, one per "group" of siteaccess.
The handler set the "right" path of CacheFile in https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/private/classes/ezpkernelweb.php#L259

Opencontent:
To further explain the issue: in a multisite (multidatabase) environment what happens is that, when an editor of website A clears the cache, the system writes the global expiry.php. That "invalidates" the cache for website B, C, D and so on, issuing a massive cache regeneration. 
The patch basically sets an expiry.php for each website.
Hope it is clearer now.

gggeek:
@Opencontent but are there any caches which are in fact supposed to be cleared across all sites, which do not get cleared any more because of your change? (sorry, did not look at the code, only thinking about the description)

crevillo:
i have to agree with @gggeek . at least this should be configured somehow. Maybe someone wants indeed clear all of the sites cache. i can think in, at least, one our multisites projects needing this...

francescor:
Yes, sure there are caches that needs to be cleared across all sites, but this operation should  be triggered by the developer (& not done that often), shouldn't be? : a great advantage (hard disk performance-wise, especially) is sure gained if editors of site A cannot clear cache of site B.

Opencontent:
That's right @gggeek @crevillo 
Added a site.ini setting to control the behavior and removed the return statement

PR #1241, opened 2016-05-19
Change value's key of Sorting in the array of parameters to fix sorting with ezfind

imadkaf:
In eZ Publish 5, when we search content using that (from SF2 action) [ eZFunctionHandler::execute('content','search', array(...)) ] , I detect that currently we cannot search by sort parameter.

**Why ?**

Because in eZFind's extension on this file extension/ezfind/classes/ezfezpsolrquerybuilder.php at line [878 ](https://github.com/ezsystems/ezfind/blob/master/classes/ezfezpsolrquerybuilder.php#L878)exactly, this key is known by 'SortBy' and not by 'SortArray',  that's why search by sorting does not work. 

**Solution :**

I just changed the value's key of Sorting in the array of parameters and checked if there is some regressions.

**2nd Solution (if 1st is bad) :**

Change in the other side on this file extension/ezfind/classes/ezfezpsolrquerybuilder.php the value's key of sorting.

### DIFF
--- a/kernel/content/ezcontentfunctioncollection.php
+++ b/kernel/content/ezcontentfunctioncollection.php
@@ -416,7 +416,7 @@ class eZContentFunctionCollection
         if ( $publishDate !== false )
             $parameters['SearchDate'] = $publishDate;
         if ( $sortArray !== false )
-            $parameters['SortArray'] = $sortArray;
+            $parameters['SortBy'] = $sortArray;
         $parameters['SearchLimit'] = $limit;
         $parameters['SearchOffset'] = $offset;
         $parameters['IgnoreVisibility'] = $ignoreVisibility;
--- a/kernel/setup/steps/ezstep_welcome.php
+++ b/kernel/setup/steps/ezstep_welcome.php
@@ -64,6 +64,11 @@ class eZStepWelcome extends eZStepInstaller
         $this->PersistenceList['tests_run'] = $testsRun;
         $this->PersistenceList['optional_tests_run'] = $testsRun;
 
+        if ( $this->hasKickstartData() )
+        {
+            return $this->kickstartContinueNextStep();
+        }
+
         return false; // Always show welcome message
     }
 


yannickroger:
Can you add a description to explain what the goal of your patch does ?

imadkaf:
Description added, you can check it.

andrerom:
This seems to be wrong change afaik.
Built-in search engine expects SortArray: https://github.com/ezsystems/ezpublish-legacy/blob/3429aa91bba018f518fcb41757c9d627a0a73c0f/kernel/search/plugins/ezsearchengine/ezsearchengine.php#L544

So I would assume ezfind needs a change here, either supporting both and take this first as it would be the right one as far as I see.

But it's a major mystery why no one has reported this before.. ping @paulborgermans @gggeek @yannickroger @peterkeung @joaoinacio

gggeek:
I kinda remember having seen this bug before, but can not find a ticket for that...

I agree with @andrerom about the fix having to be the other way around.

As for why has it not be reported as a blocker before: maybe most code which uses ezfind does not do it via fetchContentSearch, but via direct calls to eZSearch::search, and pass in the good parameter depending on the search engine in use, which generally does not change within a single project?
The only the part of the code which realistically does make use of the fetch function fetchContentSearch are the templates for the admin interface, which do not allow the user to set the sort order anyway. Maybe if they did, the bug would show.

andrerom:
That is probably the case @gggeek, somewhat what I expected as well.

paulborgermans:
Agree with @andrerom , ez find needs to be patched taking into account both variations, not the kernel

And as @gggeek pointed out, usually projects that use ezfind, just call it directly as it has many more parameters

peterkeung:
Example of how this was fixed specifically for eZ Find: https://github.com/ezsystems/ezfind/pull/78

And yes, on sites with eZ Find we just use the ezfind/search fetch function directly.

But also here is some generic functionality where we support both syntaxes:

https://github.com/ezsystems/ezpublish-legacy/blob/master/extension/ezjscore/classes/ezjscserverfunctionsjs.php#L441

Is there harm to pass both through? Note of course that the default SortArray is an array, whereby eZ Find's SortBy is a hash / associative array, so the developer needs to change their code to swap anyway.

andrerom:
Looking at the link from @peterkeung I think we should close this as won't fix, given the content format of the parameter is different, I would say it is expected that you pass the one you need.

gggeek:
@andrerom maybe we could document it better then ?

andrerom:
> @andrerom maybe we could document it better then ?

yes, PR on that would be helpful for us all.

PR #1244, opened 2016-06-18
CSS Compressing was ignoring @2x image path

skrosoft:
image path like background-image: url('../images/sample@2x.png'); wasnt replaced during CSS Compressing


andrerom:
+1

yannickroger:
@skrosoft Can you please create a JIRA issue with description and step to reproduce. We need it for backport and QA processes.

PR #1245, opened 2016-06-27
EZS-871: check if search engine plugin methods are callable

joaoinacio:
Test existence of  ezpSearchEngine methods using  `is_callable()` Instead of `method_exists()`.
This allows for automagic proxy methods (`__call()`), while also checking for the correct accessibility (public methods)

See https://jira.ez.no/browse/EZS-821

### DIFF
--- a/kernel/classes/ezsearch.php
+++ b/kernel/classes/ezsearch.php
@@ -156,7 +156,7 @@ class eZSearch
         if ( $searchEngine instanceof ezpSearchEngine )
         {
             // This method was renamed in pre 3.5 trunk
-            if ( method_exists( $searchEngine, 'supportedSearchTypes' ) )
+            if ( is_callable( array( $searchEngine, 'supportedSearchTypes' ) ) )
             {
                 $searchTypesDefinition = $searchEngine->supportedSearchTypes();  // new and correct
             }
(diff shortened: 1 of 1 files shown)


joaoinacio:
`is_callable` will return true if a magic `__call()` function exists and is public.

This should be fine as search engines shouldn't really use it blindly (or at all), in any case for reco a `method_exists()` checks is still being done on the actual searchengine

andrerom:
But then there is still some issue with this.

If the underlying search engine, either by us or by partner, is missing one of these methods, when using recommendation engine these calls will be forwarded to Reco and potentially cause issues. Code in recommendation will need to have some default value to give on `supportedSearchTypes` for instance, and return `false` for anything else when method does not exists on inner engine.

Secondly there might be code out there already using `method_exists` against search engines, that won't work when Reco engine is injected anymore.

My point is I'm really not a fan of doing such a change in a mature product that is not supposed to not break bc, even if chances are slim someone might or might not use this.

wdyt @bdunogier ?

joaoinacio:
> Secondly there might be code out there already using method_exists against search engines, that won't work when Reco engine is injected anymore.

Haven't thought about this, it would cause methods to be ignored yes.
If the chances for BC are real and not something we can simply document than it should be easy enough to take the easy route and just implement everything.

I also thought about using an Interface for a few moments to prevent these issues, but it could also cause BC break no?

andrerom:
> I also thought about using an Interface for a few moments to prevent these issues, but it could also cause BC break no?

yes, no point in that in patch release.

joaoinacio:
Thanks, closing

PR #1246, opened 2016-07-11
EZP-25771: Add handler constructor existence verification to ezextension getHandler()

miguelcleverti:
JIRA: http://jira.ez.no/browse/EZP-25771

Then getHandlerClass method instantiates the handler using Reflector method newInstanceArgs() which causes an error is the class does not have a constructor.

This fixes the issue by calling Reflector method nweInstance() if the class does not have a constructor.

### DIFF
--- a/lib/ezutils/classes/ezextension.php
+++ b/lib/ezutils/classes/ezextension.php
@@ -544,10 +544,10 @@ class eZExtension
             // we rely on the autoload system here
             if ( class_exists( $handler ) )
             {
-                // only use reflection if we have params to avoid exception on objects withouth constructor
-                if ( $handlerParams !== null && is_array( $handlerParams ) && count( $handlerParams ) > 0 )
+                $reflection = new ReflectionClass( $handler );
+                // only pass parameters if the class has a constructor
+                if ( $reflection->getConstructor() !== null && $handlerParams !== null && is_array( $handlerParams ) && count( $handlerParams ) > 0 )
                 {
-                    $reflection = new ReflectionClass( $handler );
                     $object = $reflection->newInstanceArgs( $handlerParams );
                 }
                 else


lib/ezutils/classes/ezextension.php
```diff
@@ -547,8 +547,13 @@ public static function getHandlerClass( ezpExtensionOptions $options )
                 // only use reflection if we have params to avoid exception on objects withouth constructor
                 if ( $handlerParams !== null && is_array( $handlerParams ) && count( $handlerParams ) > 0 )
                 {
-                    $reflection = new ReflectionClass( $handler );
```
joaoinacio: this shouldn't be removed

lib/ezutils/classes/ezextension.php
```diff
@@ -547,8 +547,13 @@ public static function getHandlerClass( ezpExtensionOptions $options )
                 // only use reflection if we have params to avoid exception on objects withouth constructor
```
joaoinacio: this should probably be removed, as it's ... wrong:
using `newInstanceArgs()` with parameters on classes without a constructor will result on a fatal error.

lib/ezutils/classes/ezextension.php
```diff
@@ -547,8 +547,13 @@ public static function getHandlerClass( ezpExtensionOptions $options )
...
                 if ( $handlerParams !== null && is_array( $handlerParams ) && count( $handlerParams ) > 0 )
                 {
-                    $reflection = new ReflectionClass( $handler );
-                    $object = $reflection->newInstanceArgs( $handlerParams );
+                    // if the class has no constructor do not pass parameters
```
joaoinacio: nitpick: comment is right, but in reverse... maybe  `// only pass parameters if the class has a constructor`

lib/ezutils/classes/ezextension.php
```diff
@@ -547,8 +547,13 @@ public static function getHandlerClass( ezpExtensionOptions $options )
...
+                    if ($reflection->getConstructor() != null)
+                    {
+                        $object = $reflection->newInstanceArgs( $handlerParams );
+                    } else {
+                        $object = $reflection->newInstance();
```
joaoinacio: This is basically the same as the alternative `$object = new $handler();` bellow, perhaps refactor a bit to avoid two separate if / else ?

lib/ezutils/classes/ezextension.php
```diff
@@ -544,10 +544,10 @@ public static function getHandlerClass( ezpExtensionOptions $options )
             // we rely on the autoload system here
             if ( class_exists( $handler ) )
             {
-                // only use reflection if we have params to avoid exception on objects withouth constructor
-                if ( $handlerParams !== null && is_array( $handlerParams ) && count( $handlerParams ) > 0 )
+                $reflection = new ReflectionClass( $handler );
+                // only pass parameters if the class has a constructor
+                if ( $reflection->getConstructor() !== null && $handlerParams !== null && is_array( $handlerParams ) && count( $handlerParams ) > 0 )
                 {
-                    $reflection = new ReflectionClass( $handler );
```
andrerom: I think we need to make user aware of this so it does not silently allow people to forget constructor, could you try something like:
```php
                {
                    $reflection = new ReflectionClass( $handler );
                    if ( $reflection->getConstructor() !== null )
                    {
                        $object = $reflection->newInstanceArgs( $handlerParams );
                    }
                    else
                    {
                        eZDebug::writeWarning( 'Constructor is missing but parameters are provided to class ' . $handler . " as defined in setting $iniFile [$iniSection] $iniVariable", __METHOD__ );
                        $object = new $handler();
                    }
```

joaoinacio: Since all the handlers will be called with the new injected parameter(s) with https://github.com/ezsystems/ezflow/pull/75 , it is a form of a BC break, so handlers could to be updated.
OTOH from a dev POV, why need to add an empty constructor to a handler that was working as intended?
This is mostly an issue in the 5.3 / 5.4 backports...

A warning seems like a safe choice, however since there are no constructors yet in existing handlers there may be a lot of warnings, updating handlers in flow is highly recommended too.

andrerom: It could be notice or debug as well.
Point was to make sure developers are made aware of missing constructors by mistake, which before this PR would give you a clear `Uncaught PHP Exception `, but now it is silenced.

joaoinacio: Sounds good, I'd probably add a BC doc notheless to note ezflow passes params to handler's constructor.

andrerom:
Applied the nitpicks and opened PR with that here: https://github.com/ezsystems/ezpublish-legacy/pull/1247

PR #1248, opened 2016-07-19
Ezoe ie fix

pkamps:
It fixes an issue in IE. If you insert a table into a ezxmltext attribute, in IE all table cells are collapsed making it hard to select a specific cell. With this patch each cell should be at least 20px high.


pkamps:
sorry, stupid me took the wrong base for the pull request.

PR #1249, opened 2016-07-22
Fix EZP-26096: Custom tags' text line parameters not saved correctly

dpobel:
JIRA: https://jira.ez.no/browse/EZP-26096

# Description

This patch fixes an issue in ezxmltext input parser which breaks when an attribute contains a `>`. Basically, to correctly detect an opening tag it's not enough to search for the next `>` after a `<` because the next `>` can be inside an attribute.

# Tests

manual + unit tests

### DIFF
--- a/kernel/classes/datatypes/ezxmltext/ezxmlinputparser.php
+++ b/kernel/classes/datatypes/ezxmltext/ezxmlinputparser.php
@@ -413,7 +413,7 @@ class eZXMLInputParser
         //  Regular tag: get tag's name and attributes.
         else
         {
-            $tagEndPos = strpos( $data, '>', $tagBeginPos );
+            $tagEndPos = $this->findEndOpeningTagPosition( $data, $tagBeginPos );
             if ( $tagEndPos === false )
             {
                 $pos = $tagBeginPos + 1;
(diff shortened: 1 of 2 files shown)


gggeek:
Are you sure that '<' is valid inside an attribute value ?

https://www.w3.org/TR/xml/#NT-AttValue

I says that it is definitely not valid...

dpobel:
@gggeek you are right `<` is not valid in an XML attribute but the issue here is about `>`

gggeek:
@dpobel I knew I must have been stupid! :-D

dpobel:
(trying to figure out why the tests are failing on TravisCI and not locally)

yannickroger:
+1 and thanks for the headache :sweat_smile:

glye:
+1. Doesn't look blazingly fast, but there can't be *that* many `>`'s in attribute values out there.

pkamps:
Doesn't seem to work with given text for a custom attribute:

<iframe width="100%" height="450" scrolling="no" frameborder="no" src="https://w.soundcloud.com/player/?url=https%3A//api.soundcloud.com/tracks/114399026&amp;auto_play=false&amp;hide_related=false&amp;show_comments=true&amp;show_user=true&amp;show_reposts=false&amp;visual=true"></iframe>

After saving the object version, the custom tag is broken.

peterkeung:
@dpobel Has this been tested with Philipp's fuller example above?

kernel/classes/datatypes/ezxmltext/ezxmlinputparser.php
```diff
@@ -583,6 +583,59 @@ function parseTag( &$data, &$pos, &$parent )
...
+        if ( strpos( $tagCode, '=' ) === false )
+        {
+            // this tag has no attribute, so the next '>' is the right one.
+            return $endPos;
+        }
+        if ( $this->isValidXmlTag( $tagCode ) )
+        {
+            return $endPos;
+        }
+        return $this->findEndOpeningTagPosition( $data, $tagBeginPos, $endPos + 1 );
```
pkamps: You are mixing up the offset with the absolute position. The code should read:
return $this->findEndOpeningTagPosition( $data, $tagBeginPos, $endPos - $tagBeginPos + 1 );

PR #1250, opened 2016-07-25
Fix EZP-26108: Typing right after adding a block custom tag breaks the contents

dpobel:
JIRA: https://jira.ez.no/browse/EZP-26108

# Description

Before this patch, if you type something right after adding a block custom tag, the content of the ezxmltext attribute is messed up. For instance, in Chrome, the following paragraph is added to the custom tag content, in Firefox, the custom tag is completely removed. This is happening because the newly added custom tag was completely selected, with the patch only its content is selected.

# Tests

manual tests


joaoinacio:
The TinyMCE docs the `select()` method and params is a bit weird,
> | content | Boolean | Optional bool state if the contents should be selected or not on non IE browser.

So is the fix OK for IE/Edge as well (or was there no issue in the first place) ?

dpobel:
according to 2nd line, the bug occurred only in Chrome and Firefox (under different form) and with the fix Edge is still fine.

yannickroger:
+1

andrerom:
+1 assuming this does not cause problems for embed blocks

PR #1251, opened 2016-07-26
EZP-26058: update script for non-breaking spaces in ezxmltext

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-26058

### DIFF
--- a/update/common/scripts/updatenbxmlcontents.php
+++ b/update/common/scripts/updatenbxmlcontents.php
@@ -44,7 +44,7 @@ $script->initialize();
 $db = eZDB::instance();
 
 if ( $optDryRun )
-
+{
     $cli->warning( "dry-run mode" );
 }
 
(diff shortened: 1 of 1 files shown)


andrerom:
+0.5

ping @glye / @yannickroger / @dpobel

glye:
+1

Deja Vu - haven't we reviewed this before? :) (The inline fix in kernel/oe maybe.)

I thought the Elvis operator `?:` was still problematic for backports, but it seems even 4.4 requires PHP 5.3, so it should be fine.

joaoinacio:
> I thought the Elvis operator ?: was still problematic for backports, but it seems even 4.4 requires PHP 5.3, so it should be fine.

tbh I copied-pasted that section from another script, but could update it anyway just in case.

update/common/scripts/updatenbxmlcontents.php
```diff
@@ -0,0 +1,118 @@
...
+$limit = array(
+    "offset" => 0,
+    "limit" => $optIterationLimit,
+);
+
+$script->initialize();
+$db = eZDB::instance();
+
+if ( $optDryRun )
+
```
glye: `{` missing ;)

joaoinacio:
updated according to @dpobel 's feedback about other tags:
> don't we need to apply the same process in others tag (like heading for instance) ?

AFAICT, heading is the only other tag for which the problem applies.
In any case the tags (`paragraph`, `header`) are now defined in an array first, the regex is only slightly more complex by using named match; see commit https://github.com/ezsystems/ezpublish-legacy/pull/1251/commits/25bac64394e96f22def07fd6d634fb922e6edc3a

andrerom:
+1

update/common/scripts/updatenbxmlcontents.php
```diff
@@ -0,0 +1,120 @@
...
+    do {
+        $xmlData = preg_replace( $pattern, $replace, $xmlData, -1, $countReplaced );
+    } while ($countReplaced > 0);
+
+    if ( $verbose ) {
+        eZCLI::instance()->output( "Updating data for content #$contentId (ver. $version) ..." );
+    }
+    if ( !$optDryRun ) {
+        eZDB::instance()->query( "UPDATE ezcontentobject_attribute SET data_text='$xmlData' WHERE id='$id'" );
```
wizhippo: this fails if $xmldta contains a ' character.

PR #1163, opened 2015-04-28
Fix EZP-19376: CSS packer and font-face conflict

sterrien:
https://jira.ez.no/browse/EZP-19376

This PR changes the URL replacement algorithm. **The new algorithm avoids the following problem.**

When a file contained this kind of url :

``` css
  src: url("fonts/HelveticaNeue55-Roman.eot");
  src: url("fonts/HelveticaNeue55-Roman.eot?#iefix") format("embedded-opentype");
```

we had these lines in return :

``` css
  src: url("/extension/my_extension/design/my_design/fonts/HelveticaNeue55-Roman.eot");
  src: url("/extension/my_extension/design/my_design//extension/my_extension/design/my_design/fonts/HelveticaNeue55-Roman.eot?#iefix");
```

PR #1252, opened 2016-07-29
Fix EZP-19376: ezjscore CSS packer duplicate url path replacement

joaoinacio:
https://jira.ez.no/browse/EZP-19376

This makes sure that urls in CSS are not replaced with paths "recursively", see original description:

> When a file contained this kind of url :
> ```
>   src: url("fonts/HelveticaNeue55-Roman.eot");
>   src: url("fonts/HelveticaNeue55-Roman.eot?#iefix") format("embedded-opentype");
> ```
> we had these lines in return :
> ```
>   src: url("/extension/my_extension/design/my_design/fonts/HelveticaNeue55-Roman.eot");
>   src: url("/extension/my_extension/design/my_design//extension/my_extension/design/my_design/fonts/HelveticaNeue55-Roman.eot?#iefix");
> ```

Alternative for https://github.com/ezsystems/ezpublish-legacy/pull/1163 , which addresses the complexity of the initially proposed fix.

### DIFF
--- a/extension/ezjscore/classes/ezjscpacker.php
+++ b/extension/ezjscore/classes/ezjscpacker.php
@@ -495,6 +495,7 @@ class ezjscPacker
     {
         if ( preg_match_all( "/url\(\s*[\'|\"]?([A-Za-z0-9_\-\/\.\\%?&#=]+)[\'|\"]?\s*\)/ix", $fileContent, $urlMatches ) )
         {
+           $urlPaths = array();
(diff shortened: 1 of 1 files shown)


joaoinacio:
ping @andrerom  @yannickroger

andrerom:
+1

joaoinacio:
ping @glye

glye:
+1, good solution.

PR #1254, opened 2016-08-12
Fix EZP-26171: ezoe php7

wizhippo:
ezoe breaks under php7 due to the constructor using a reference.

### DIFF
--- a/extension/ezoe/ezxmltext/handlers/input/ezoexmlinput.php
+++ b/extension/ezoe/ezxmltext/handlers/input/ezoexmlinput.php
@@ -46,7 +46,7 @@ class eZOEXMLInput extends eZXMLInputHandler
      * @param string $aliasedType
      * @param eZContentObjectAttribute $contentObjectAttribute
      */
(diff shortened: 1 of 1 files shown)


wizhippo:
Alternative approach would be to fix ezextension to handle passing by reference.

antoligy:
Looks good!

joaoinacio:
Note: seems like a duplicate of EZP-26028, problem/fix should apply to all input/output handlers (see ezxmltext datatype).

wizhippo:
This fix is not needed if https://github.com/ezsystems/ezpublish-legacy/pull/1255 is merged as it handles the passing of references.

andrerom:
Closed by #1259

andrerom:
@wizhippo Thanks for pushing this topic btw, as you might have seen we found several other places it was mistakenly sending by reference so jumped at fixing it all over.

wizhippo:
@andrerom Thanks for finding the others and getting the fix merged.

PR #1256, opened 2016-08-23
Fix EZP-26096: additional fix to handle several attributes with special characters

dpobel:
JIRA: https://jira.ez.no/browse/EZP-26096

# Description

This is an additional fix with the corresponding unit test for https://github.com/ezsystems/ezpublish-legacy/pull/1249 provided by @pkamps in https://github.com/ezsystems/ezpublish-legacy/pull/1249#discussion_r73696903 (thanks :)) because the original fix was only working if there was only one attribute with a `>`.

# Tests

manual tests + unit tests


andrerom:
+1

peterkeung:
+1 (we have been using this for a client)

yannickroger:
+1

PR #1257, opened 2016-08-26
Fix EZP-25089: Add ezpEvent on object state assignment

peterkeung:
https://jira.ez.no/browse/EZP-25089

Before this patch, you would end up in a scenario where an object state was changed from being not available to available but the front-end would crash with an error related to the old state.

Needs to be combined with https://github.com/ezsystems/LegacyBridge/pull/61

### DIFF
--- a/doc/features/5.4/event.txt
+++ b/doc/features/5.4/event.txt
@@ -2,6 +2,7 @@ Event system
 ============
 
 - Added new events for image aliases removal
+- 5.4.6 Added new events for state assigments
 
 New events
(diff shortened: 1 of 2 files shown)


peterkeung:
Ping @andrerom, the original author of this patch

peterkeung:
Attribution added to @andrerom

peterkeung:
Ping @thiagocamposviana who originally discovered the bug with me

andrerom:
+1

PR #1258, opened 2016-08-31
Fix EZP-26224: escape xml data string in update script

joaoinacio:
Jira: https://jira.ez.no/browse/EZP-26224


glye:
Looks like +1

yannickroger:
+1

PR #1261, opened 2016-09-14
Fix EZP-26300: Fatal error when going through defaultExceptionHandler with PHP7

lolautruche:
> https://jira.ez.no/browse/EZP-26300

With PHP 7, exception handlers can now receive `Error` exceptions. Those exceptions are internal to PHP and are not directly linked to `Exception` class, except that they both implement `Throwable` interface, which is present in PHP 7 only.

This patch just removes `Exception` type hint in `eZExecution::defaultExceptionHandler()` in order to avoid fatal errors when receiving the new `Error` exceptions.

### DIFF
--- a/lib/ezutils/classes/ezexecution.php
+++ b/lib/ezutils/classes/ezexecution.php
@@ -177,7 +177,7 @@ class eZExecution
      * @params Exception the exception
      * @return void
      */
-    static public function defaultExceptionHandler( Exception $e )
+    static public function defaultExceptionHandler( $e )
     {
         if( PHP_SAPI != 'cli' )
         {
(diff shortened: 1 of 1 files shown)


lolautruche:
Ping @andrerom @bdunogier @yannickroger

bdunogier:
Looks good to me, +1.

@andrerom / @glye ?

glye:
+1. (Nice to know.)

lib/ezutils/classes/ezexecution.php
```diff
@@ -177,7 +177,7 @@ static public function registerShutdownHandler( $documentRoot = false )
      * @params Exception the exception
```
andrerom: Please update the description to hint of the different types we get here now.

andrerom:
+1 besides the comment

lolautruche:
Done :-)

lolautruche:
Thanks for merging so quickly 😄

PR #1262, opened 2016-09-20
Better contentobjects cleanup

crafics:


composer.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name":        "ezsystems/ezpublish-legacy",
+    "name":        "sport1online/ezpublish-legacy",
```
Tantacom: mmm, sorry, i don't think ezsystems would like to change this :)
even now i realized this pr is closed... is it closed in favour or any other?

crafics: sorry.. I accidentally pushed the PR Button on github without changing the master to sport1 :) pls ignore it

PR #1265, opened 2016-10-17
Fix EZP-26449: Wrong version translation list in eZContentOperationCollection

lolautruche:
> https://jira.ez.no/browse/EZP-26449

When publishing a content from legacy (e.g. from admin interface), `content/publish` operation copies missing translations on the published version.
However, `eZContentOperationCollection::copyTranslations()` ignores the version language mask when retrieving the translation list for the version to publish, using `$version->translationList( false, false )`. While this is usually fine as a translation is generally a full bunch of translated attributes, this can lead to unexpected behavior, especially when using a mix of legacy and platform scripts.

Typical example is when using a _wait until date_ workflow (legacy admin/kernel), combined with several scripts from Symfony stack, like migration scripts for example:
- Webmaster publishes from admin interface with a publication date set in the future. Content has several translations.
- DevOps deliver a new version of code with migration scripts to update ContentTypes, including the one used by the webmaster, to add 1 new field
- Workflow cronjob is run to publish content from webmaster

The result here will be:
- Translation published by the webmaster will be OK
- All other translations will be broken, only having the new field.

This is due to `$version->translationList( false, false )` used in `eZContentOperationCollection::copyTranslations`, called by the workflow cronjob. This method will list all available translations for given version, by calling a `SELECT DISTINCT language_code FROM ezcontentobject_attribute`. As we added a new field using the public API, all translations are taken into account, instead of only published language.

Use `language_mask` from version instead of `$version->translationList()`.

### DIFF
--- a/kernel/content/ezcontentoperationcollection.php
+++ b/kernel/content/ezcontentoperationcollection.php
@@ -651,18 +651,18 @@ class eZContentOperationCollection
         $publishedLanguageCodes = array_keys( $publishedLanguages );
 
         $version = $object->version( $versionNum );
-        $versionTranslationList = $version->translationList( false, false );
+        $versionTranslationList = array_keys( eZContentLanguage::languagesByMask( $version->attribute( 'language_mask' ) ) );
 
         foreach ( $publishedVersionTranslations as $translation )
         {
-            if ( in_array( $translation->attribute( 'language_code' ), $versionTranslationList )
-              || !in_array( $translation->attribute( 'language_code' ), $publishedLanguageCodes ) )
+            $translationLanguageCode = $translation->attribute( 'language_code' );
+            if ( in_array( $translationLanguageCode, $versionTranslationList )
+                || !in_array( $translationLanguageCode, $publishedLanguageCodes ) )
             {
                 continue;
             }
 
-            $contentObjectAttributes = $translation->objectAttributes();
-            foreach ( $contentObjectAttributes as $attribute )
+            foreach ( $translation->objectAttributes() as $attribute )
             {
                 $clonedAttribute = $attribute->cloneContentObjectAttribute( $versionNum, $publishedVersionNum, $objectID );
                 $clonedAttribute->sync();


lolautruche:
Review ping @andrerom @bdunogier @yannickroger @glye

kernel/content/ezcontentoperationcollection.php
```diff
@@ -651,18 +651,18 @@ static public function copyTranslations( $objectID, $versionNum )
...
 
         $version = $object->version( $versionNum );
-        $versionTranslationList = $version->translationList( false, false );
+        $versionTranslationList = array_keys( eZContentLanguage::languagesByMask( $version->attribute( 'language_mask' ) ) );
 
         foreach ( $publishedVersionTranslations as $translation )
         {
-            if ( in_array( $translation->attribute( 'language_code' ), $versionTranslationList )
-              || !in_array( $translation->attribute( 'language_code' ), $publishedLanguageCodes ) )
+            $translationLanguageCode = $translation->attribute('language_code');
```
glye: Tiny nitpick: Legacy CS (spaces within parens)

lolautruche: Argh thanks :-)

lolautruche: Fixed

joaoinacio:
My 2 cents:
IIUC, the only real change is at the L654, correct?  (perhaps the other refactoring could be ommited?)
```diff
-        $versionTranslationList = $version->translationList( false, false );
+        $versionTranslationList = array_keys( eZContentLanguage::languagesByMask( $version->attribute( 'language_mask' ) ) );
```

OTOH, It seems that the way `translationList()` works could potentially cause different issues with similar use case(s) of using SF scripts to update Types/Fields.

For example, a quick look shows that calling `updateLanguageMask()` function with default parameters would update the `language_mask` to those translations.

I wonder if the issue is rather on the ContentType update?

lolautruche:
> IIUC, the only real change is at the L654, correct? (perhaps the other refactoring could be ommited?)

That's correct. I can split into 2 commits if you prefer.

> OTOH, It seems that the way translationList() works could potentially cause different issues with similar use case(s) of using SF scripts to update Types/Fields, in which case more severe problems could happen (and this fix would become ineffective)
>
> For example, a quick look shows that calling updateLanguageMask() function with default parameters would update the language_mask to those translations.

AFAICT it is perfectly safe. The problem with `$version->translationList()` is that it looks into **all** published languages for **all** attributes in the current version (including the new one). But as there is a new field, it then considers that the version has been saved for all those languages, which is not correct. The only **reliable** information for this is the language mask stored in the version, hence my fix.
It is perfectly safe as it relies on the edit language mask for the version to be published. I don't see any flaw here.

Yes `updateLanguageMask()` is called because all translations are copied into the version to be published, and thus the information must be updated.

> I wonder if the issue is rather on the ContentType update?

That's what I thought in the beginning, but that's not the case. Looking at how ContentType handler works (in the persistence layer), it's perfectly consistent with how the API/SPI is working. The glitch comes from a mixed usage of API and legacy. And part of the code affected by my patch was actually mis-used, but OK with how legacy works.

yannickroger:
Thanks for the fix @lolautruche sorry about the delay for merging.

pkamps:
So if it's not save to rely on `$version->translationList()` - shouldn't we also update `function updateLanguageMask()` because it also relies on `$version->translationList()`.

Or even change `function translationList()` to rely on `$version->attribute( 'language_mask' )`

Also, I still don't fully understand why a migration script is causing `function translationList()` to return unwanted language codes for a 'pending' version (_wait until date_ workflow).

PR #1268, opened 2016-11-25
EZP-26209 Textline legacy search indexing causes TransactionError

wawrzynkiewicz:
Causes SQL error: #1264 - Out of range value for column 'integer_value' at row 1
@integerValue generation the value needs to be verified against MySQL's max int value.

### DIFF
--- a/kernel/search/plugins/ezsearchengine/ezsearchengine.php
+++ b/kernel/search/plugins/ezsearchengine/ezsearchengine.php
@@ -96,7 +96,11 @@ class eZSearchEngine implements ezpSearchEngine
                     // Split text on whitespace
                     if ( is_numeric( trim( $text ) ) )
                     {
-                        $integerValue = (int) $text;
+                        //verify if value does not exceed MySQL int(11) unsigned maximum value
+                        if( (int)$text > (int)2147483647 )
+                            $integerValue = 0;
+                        else
+                            $integerValue = (int) $text;                      
                     }
                     else
                     {


yannickroger:
ping @joaoinacio 
Also keep in mind that we also support PGsql.

blankse:
+1 I have the same error

pkamps:
I run into the same issue.

kernel/search/plugins/ezsearchengine/ezsearchengine.php
```diff
@@ -96,7 +96,11 @@ public function addObject( $contentObject, $commit = true )
                     // Split text on whitespace
                     if ( is_numeric( trim( $text ) ) )
                     {
-                        $integerValue = (int) $text;
+                        //verify if value does not exceed MySQL int(11) unsigned maximum value
```
adamwojs: Based on magic number one line below, I think that you have on your mind the maximum **signed** value of `int`. MySQL documentation says (https://dev.mysql.com/doc/refman/5.7/en/integer-types.html) that maximum unsigned value of int is `4294967295` 😉

peterkeung:
What is the "integer_value" field used for anyway? That would determine whether setting it to "0" (rather than the max value, for example) is an appropriate action.

blankse:
It is possible to define a searchtype.
https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/search/plugins/ezsearchengine/ezsearchengine.php#L1356

In the types integer, integers, byrange, byidentifierrange, integersbyidentifier the integer_value is used.

I would prefer the changes from @pkamps in his fork :)

andrerom:
@blankse Which changes do you refer to?

blankse:
@andrerom Yes, the linked one.

It set the max value and takes the min value into account.

andrerom:
@blankse with only nitpic of not referring just to mysql on the constants, I agree, it's cleaner approach. Maybe something that can be opened here?

andrerom:
Closing, handled in #1338

PR #1269, opened 2016-11-25
EZP-26680 INI Cache files are generated with wrong filepaths

wawrzynkiewicz:
When publicAPI calls are executed via Symfony controller and for example a user is created via UserService, this Signal is triggered into legacy System cache management.

The eZINI::instance() is called inside user datatype which checks site.ini cache file. In the cache file are relative paths to override ini files which do not work because application is in Symfony context (and not cwd in legacy).

The override file is not found via file_exists() and the new generated cachefile is w/o override files cached, which "destroys" legacy application.

### DIFF
--- a/lib/ezutils/classes/ezini.php
+++ b/lib/ezutils/classes/ezini.php
@@ -482,26 +482,26 @@ class eZINI
                  else
                     $overrideFile = eZDir::path( array( $rootDir, $overrideDir, $fileName ) );
 
-                if ( file_exists( $overrideFile . '.php' ) )
+                if ( file_exists( __DIR__ . '/../../../' . $overrideFile . '.php' ) )
                 {
                     // recursion eZDebug::writeStrict( "INI files with *.ini.php suffix is DEPRECATED, use *.ini or *.ini.append.php instead: $overrideFile.php", __METHOD__ );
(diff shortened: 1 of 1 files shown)


andrerom:
-1 Might be missing legacy fallback call which would make sure folder is right for legacy?

gggeek:
I was thinking along the same lines...

wawrzynkiewicz:
The requested ini file is always stored absolute the additional one's relative. So there is a inconsistency on this...

<?php
// This is a auto generated ini cache file, time created:Fri, 25 Nov 16 12:23:32 +0100
$data = array(
'rev' => 2,
'created' => '2016-11-25T12:23:32+01:00',
'charset' => "utf-8",
'files' => array (0 => '/www/example/ezpublish_legacy/lib/ezutils/classes/../../../settings/site.ini',1 => 'settings/override/site.ini.append.php',
),
'file' => '/www/example/ezpublish_legacy/lib/ezutils/classes/../../../settings/site.ini',
'val' => array...

joaoinacio:
IIUC,`eZDir::path()` should return an absolute path derived from `[$rootDir, $overrideDir, $fileName]`.
If that is not the case, the issue could be elsewhere?

PR #1270, opened 2016-11-25
EZP-15983 - Only include ezp_override when present

antoligy:
When generating autoloads on a fresh ezpublish-legacy install through Symfony, the missing file raises a notice which is caught by the Symfony exception handler - This means that the autoloads can't be generated through Symfony, unless they have previously been generated by legacy.

This change is trivial and mitigates the issue.

Duplicate/response to https://jira.ez.no/browse/EZP-15983 - Please don't hate me for using a file_exists @bdunogier!

### DIFF
--- a/autoload.php
+++ b/autoload.php
@@ -120,7 +120,8 @@ if ( !class_exists( 'ezpAutoloader', false ) )
                 {
                     // won't work, as eZDebug isn't initialized yet at that time
                     // eZDebug::writeError( "Kernel override is enabled, but var/autoload/ezp_override.php has not been generated\nUse bin/php/ezpgenerateautoloads.php -o", 'autoload.php' );
-                    if ( $ezpKernelOverrideClasses = include __DIR__ . '/var/autoload/ezp_override.php' )
+                    $ezpKernelOverridePath = __DIR__ . '/var/autoload/ezp_override.php';
+                    if ( file_exists( $ezpKernelOverridePath ) && $ezpKernelOverrideClasses = include $ezpKernelOverridePath )
                     {
                         self::$ezpClasses = array_merge( self::$ezpClasses, $ezpKernelOverrideClasses );
                     }


emodric:
> When generating autoloads on a fresh ezpublish-legacy install through Symfony

I think I never ran into this issue so I'm curious about what exactly do you mean by "through Symfony" :)

antoligy:
Hey, I'm talking about an install through `php app/console ezpublish:legacy:script bin/php/ezpgenerateautoloads.php` from a fresh install (I.E. after dumping autoloads in composer)

emodric:
Hm.. Okay! I still don't remember experiencing the issue :)

How about generating an empty file rather than using a `file_exists` ?

gggeek:
Cant you run ezpgenerateautoloads without going thru the sf console? That's how I generally do it...

antoligy:
Running directly works, but is unreliable as it doesn't run through the Symfony console.  Meaning it doesn't pick up Symfony includes (atm we run in prod mode to avoid the exception, but that is super hacky!)

@emodric Generating an empty file doesn't work, as the autoload generator immediately attempts to use the included file as an array!

emodric:
> @emodric Generating an empty file doesn't work, as the autoload generator immediately attempts to use the included file as an array!

So generate a file with an empty array? :)

pkamps:
I'm not sure but including a file with an empty array is likely to be slower than a `file_exists`. Then we're talking about a script that gets only executed from time to time. Why not just merge this pull request?

antoligy:
Creating an empty file feels like a band-aid over what feels like a real problem - Given the direction of the LegacyBridge to remove deprecation warnings this feels like an apt PR.

That being said, I am very open to hearing other methods of permanently solving this issue. 🙂

pkamps:
The code in question is wrapped in:

`if ( defined( 'EZP_AUTOLOAD_ALLOW_KERNEL_OVERRIDE' ) and EZP_AUTOLOAD_ALLOW_KERNEL_OVERRIDE )`

Why don't you set EZP_AUTOLOAD_ALLOW_KERNEL_OVERRIDE to _false_?

antoligy:
In this particular case, I actually attempt to generate kernel override autoloads through `ezpgenerateautoloads.php` which in turn invokes this autoloader.

joejonesbluetel:
@pkamps In our use case we have to use kernel overrides, so disabling them altogether isn't really an option for us. @antoligy's conditional is within the check that you mentioned so if they are disabled then this check is obviously never performed.

Also looking at the autoload.php file don't we already have a similar check on the ezp_extension.php file? To make sure that it exists before loading it. Presumably to correct the same problem for the extension autoload file?

emodric:
All this talk and I still haven't been able to reproduce the issue with `ezpublish:legacy:script` :) No warnings whatsoever in logs.

antoligy:
I'm able to reliably reproduce on PHP 5.6 by using a fresh installation of eZ Publish Platform which through composer populates the `ezpublish_legacy` directory.  `var/autoload/ezp_override.php` is not distributed with eZ Publish and so does not exist when this command is run (I've hooked it into the post autoload dump script run by composer, however this makes little difference as I can reproduce without), which is why an E_WARNING is raised.

Applying this fix manually by running a `patch`, then running this, solves the problem reliably.

Plopix:
I have reproduced that couple of times too.

pkamps:
Performance doesn't seem to be a problem and given that the code already contains a few `file_exists`:

+1

antoligy:
Bump? 😭

PR #1272, opened 2016-12-14
Fix EZP-26798: Query parameters get lost when being redirected with m…

blankse:
…obile device detect filter

JIRA: https://jira.ez.no/browse/EZP-26798


kernel/private/classes/ezpmobiledeviceregexpfilter.php
```diff
@@ -133,7 +133,7 @@ public function redirect()
                 }
```
yannickroger: Should this redirect be also updated?

blankse: Yes, you're right. I updated it.

blankse:
@bdunogier @andrerom Can you review it?

yannickroger:
Thanks @blankse

PR #1274, opened 2017-01-12
eZCharTransform::commandUrlCleanupIRI replaces double quotes

pkamps:
This bug came up in a client project. I'm pretty sure we want to replace double quotes in that function.


yannickroger:
Could you please create a JIRA issue with steps to reproduce? We need it to send it to our QA team before merge.

PR #1271, opened 2016-11-29
Fix sticky object relations

pkamps:
This pull request was developed here:
https://github.com/mugoweb/ezpublish-legacy/pull/31

A previous pull request introduced an issue with the object relation list. I repost the steps to reproduce the issue here. More details are on the link above:

== TESTING INSTRUCTIONS ==

1) Create a node that has a objectrelationslist attribute.
2) Add an object relation in that attribute
3) Publish the node
4) Confirm that the relation is listed on the relation tab
5) Edit the newly created node
6) Remove the object relation you previously created
7) Publish the node
8) Confirm that the relation is not listed on the relation tab

Without this pull request, you would still see the object relation listed in the relation tab.

PR #1277, opened 2017-01-20
Remove object link if user removes a related object in the objectrelationlist datatype attribute

pkamps:
The patch comes from the ez publish enterprise version. It fixes a bug in the objectrelationlist attribute: when removing an object relation in the attribute it was not removing it from the object relation DB table. This patch fixes the issue.

This patch replaces:
https://github.com/ezsystems/ezpublish-legacy/pull/1271

#1271 also contains steps how to test the fix.


gggeek:
Nitpick, but why not keep the variable names the same as they where ? It would make it easier to 'get' what has changed and what not

pkamps:
That's how it's fixed in the enterprise version. The code style is also not consistent... but there is probably a benefit to stay as close as possible to the enterprise code version.

PR #1279, opened 2017-02-15
Allow to set response headers on error pages

pkamps:
In ezp it's not possible to set custom response headers for error pages. ezp only allows you to set response headers by requested URLs (see HTTPHeaderSettings in site.ini).

This pull request allows you to specify custom response header for error pages on ezp. That's an important feature in case your site is behind a reverse proxy (CDN akamai, CloudFront, Varnish) and you control the caching TTLs with response headers.

You can test this pull request by accessing a URL that does not exists. From that request you should get a response with this header: 'Cache-Control:public, must-revalidate, max-age=300'

The only small complexity in this pull request is the fact that we re-use the 'Status' header in order to build the response status code - for example '404 Not Found'.

### DIFF
--- a/kernel/error/view.php
+++ b/kernel/error/view.php
@@ -56,39 +56,47 @@ $GLOBALS["eZRequestError"] = true;
 
     if ( $errorHandlerType != 'redirect' )
     {
-        // Set apache error headers if error.ini tells us to
+        // Set response headers if error.ini tells us to
+        // $errorType is 'kernel' or 'shop'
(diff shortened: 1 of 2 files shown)


gggeek:
bc break, should at least be documented as cuh

pkamps:
You're right. I probably can extend the pull request and still support the old settings variable 'HTTPName'. But is it worth it? In all my ezp projects I never had to change the HTTP response status message for error pages. I don't even know why that's configurable - maybe to translate it into another language?

pkamps:
Latest commit is avoiding the BC break.

PR #1280, opened 2017-02-16
Enable/disable the generation of the "Content-Length" response header

pkamps:
Allows to have ezp sending the Content-Length header - controlled by a setting.
With the Content-Length header, it is possible to enable compression (gzip) for CloudFront - compare:

http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/ServingCompressedFiles.html

I'm not sure if there are any downsides to send the Content-Length header, that's why I have that functionality controlled by a setting.


peterkeung:
JIRA issue: https://jira.ez.no/browse/EZP-27069

PR #1281, opened 2017-02-23
EZP-27121: Fix double-encoding of preview image URLs in the admin interface

pkamps:
831021dc-f766-11e6-8fcb-a364721040c3

The URL for this preview image is getting encoded twice:

encodeURI(record.getData('thumbnail_url')); in design/admin/javascript/ezajaxsubitems_datatable.js
in https://github.com/mugoweb/ezpublish-legacy/blob/master/extension/ezjscore/classes/ezjscajaxcontent.php#L317 -- transformURI does the encoding
That issue never surfaced because the image datatype in ezp is escaping special characters if it's creating the file name for the image file. So an image name "test{{dd}}test" results in a file name like this "test-dd-test". So there is never the need to encode anything for image URLs.

But this the double-encoding will cause issues if the image URL is coming from other systems like the mugo_dam which is not simplifying the image file name - and needs URL encoding (just not twice).

### DIFF
--- a/design/admin/javascript/ezajaxsubitems_datatable.js
+++ b/design/admin/javascript/ezajaxsubitems_datatable.js
@@ -60,7 +60,7 @@ var sortableSubitems = function () {
         }
 
         var thumbView = function(cell, record, column, data) {
-            var url = encodeURI(record.getData('thumbnail_url'));
+            var url = record.getData('thumbnail_url');
             if (url) {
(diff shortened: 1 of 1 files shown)


andrerom:
Issue need + udpate of PR title.

pkamps:
Issue: https://jira.ez.no/browse/EZP-27121

PR #1282, opened 2017-02-23
Add state change/assign to the audit log

thiagocamposviana:
We just ran into an issue on one of our customers where it would have been really good to know who made a state change. Thus it would be good to make this an audit log event.


gggeek:
github snafus, my bread and butter :-D

PR #1283, opened 2017-02-25
EZP-27011: Solr index not updated when changing a content priority

andrerom:
> Issue: https://jira.ez.no/browse/EZP-27011

Edited & Tested by @vnivuahc, ref: https://github.com/netgen/ezplatformsearch/issues/11

Ready for review @emodric @yannickroger @glye

### DIFF
--- a/kernel/content/ezcontentoperationcollection.php
+++ b/kernel/content/ezcontentoperationcollection.php
@@ -736,6 +736,11 @@ class eZContentOperationCollection
 
        eZContentObject::fixReverseRelations( $objectID, 'move' );
 
+        if ( !eZSearch::getEngine() instanceof eZSearchEngine )
+        {
+            eZContentOperationCollection::registerSearchObject( $objectID );
+        }
+
        return array( 'status' => true );
     }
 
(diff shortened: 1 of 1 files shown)


emodric:
I'm not sure I understand `if ( !eZSearch::getEngine() instanceof eZSearchEngine )`, but other than that, looks good :+1:

andrerom:
> I'm not sure I understand if ( !eZSearch::getEngine() instanceof eZSearchEngine ), but other 

@emodric It's basically skipping the reindex if search engine is the default on which does not need to update anything on index changes as it is using the storage data for search.

emodric:
>>> It's basically skipping the reindex if search engine is the default on which does not need to update anything on index changes as it is using the storage data for search.

But doesn't `eZSearch::getEngine()` return any currently configured search engine, [judging by the code of the method](https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/classes/ezsearch.php#L463-L467)? So if search engine from `ezplatformsearch` extension is activated, it would be returned.

andrerom:
@emodric yes and the index call will be made, like we want to in that case.

emodric:
> yes and the index call will be made, like we want to in that case.

Damn it, I misread the interface :) Custom search engines are not instances of `eZSearchEngine` but `ezpSearchEngine` :D

emodric:
@andrerom Going in, then? :)

andrerom:
It was sent for QA, there where no lables for that here but added it so it's more clear ;)

andrerom:
PR is blocked as during testing for some reason  index does not get updated on first attempt, only second time priority is updated.

If someone could verify and shed some light on to why this might happen that would be great.

andrerom:
Merging as reported says the fix solves the issue, QA/Support seems to have different setup.

PR #1285, opened 2017-03-03
Add state change/assign to the audit log

thiagocamposviana:
We just ran into an issue on one of our customers where it would have been really good to know who made a state change. Thus it would be good to make this an audit log event.

@peterkeung


brookinsconsulting:
+1

peterkeung:
JIRA issue: https://jira.ez.no/browse/EZP-27070

PR #1288, opened 2017-03-13
EZP-27101 : removing recursionProtect after fetching all metaData of …

flovntp:
…a relationList

This solve issue https://jira.ez.no/browse/EZP-27101

### DIFF
--- a/kernel/classes/datatypes/ezobjectrelationlist/ezobjectrelationlisttype.php
+++ b/kernel/classes/datatypes/ezobjectrelationlist/ezobjectrelationlisttype.php
@@ -1603,6 +1603,8 @@ class eZObjectRelationListType extends eZDataType
             $metaDataArray = array_merge( $metaDataArray, $attributeMetaDataArray );
         }
 
+        eZContentObject::recursionProtectionEnd();
+
         return $metaDataArray;
     }
 


flovntp:
my fix remove $GLOBALS["ez_content_object_recursion_protect"] after looping on all contents of a relationList.
Like this, we still avoid adding twice metadata of the same relationList but allow getting metadata in another relationList for the same content.

pkamps:
Side note: the recursion protection is avoiding infinite loops. For example, if you edit a content object and then select exactly this content object in a objectrelationlist attribute.

I'm able to reproduce the issue. It's maybe a bit of an unusual use case (2 object relation attributes both pointing to the same related object) - but it's definitely a bug.

pkamps:
There is a problem with this pull request: it is possible to end up in an infinite execution loop.
Here the steps to reproduce the issue:
- Create a class with 2 objectrelationlist attributes
- Create an instance (object) of this class
- Edit the instance and add object relations to both attributes. For both attributes select the content object you currently edit.
- Hit 'Send for publishing'

natanael89:
No longer needed because of https://github.com/ezsystems/ezpublish-legacy/pull/1372

PR #1289, opened 2017-03-29
EZP-27171: Fix escaping for custom tag attribute values

pkamps:
The ezxmltext datatype allows editors to embed custom tags. For example the 'factbox'. A custom tag may have input fields allowing the editor to type in text. For example 'Title' for the 'factbox'.

If you type in a " (double-quote) char into the custom tag field, it is getting encoded twice before the value gets stored in the database. That's problematic because the template that renders the custom tag is not receiving the " char but '&quot;' instead.

Steps to reproduce the issue:
- you have a xmltext attribute (like article body)
- in this attribute you add a custom tag
- the custom tag has an field and you add some HTML code into this field - something like "<iframe src="http://www.mugo.ca"></iframe>"
- You save the content object (Send to publishing)
- On the public site, the HTML for the custom tag is broken. It renders something like
"<iframe src&quot;http://www.mugo.ca&quot;></iframe>"

More details where the problem is:

When you hit 'Send for publishing', the HTML form will send all attribute values of the content object to the server. In case of the xmltext attributes, it sends an HTML string. In my test I added a custom tag 'factbox' - the HTML string looks similar to:
```
<div type="custom" class="ezoeItemCustomTag factbox" customattributes="title|&quot;page.html&quot;attribute_separationalign|right">
<p>factbox</p>
</div>
```
For the title field, I typed in "page.html" (with quotes). You can see that those double-quotes are escaped. That's good, otherwise it would break the HTML string (doube-quote in an HTML attribute value).

This HTML string is send to:
`kernel/classes/datatypes/exzmltext/ezxmlinputparser.php`

That class turns the HTML into an XML format. That XML string is getting stored in the database. In case of my example, that's what you get in the DB:

```
<paragraph xmlns:tmp="http://ez.no/namespaces/ezpublish3/temporary/">
    <custom name="factbox" custom:title="&amp;quot;page.html&amp;quot;" custom:align="right"/>
</paragraph>
```

Here you can see the problem: `&amp;quot;`

That " char gets encoded twice:
```
"         ->   &quot;
&quot;     -> $amp;quot;
```

There is no good reason for the double-encoding. In XML, you are allowed to have the string &quot; as an attribute value.

The double encoding happens in
`extension/ezoe/ezxmltext/handlers/input/ezoeinputparser.php Line: 274`

`$element->setAttribute( $qualifiedName, $value );`

setAttribute is escaping the given $value - see http://stackoverflow.com/questions/7294134/php-xml-dom-unwanted-escaping-characters-i-cannot-write-qnot

I changed it to
`$element->setAttribute( $qualifiedName, htmlspecialchars_decode( $value ) );`

That will avoid the double-encoding and will render the correct value in the templates.

### DIFF
--- a/extension/ezoe/ezxmltext/handlers/input/ezoeinputparser.php
+++ b/extension/ezoe/ezxmltext/handlers/input/ezoeinputparser.php
@@ -262,7 +262,7 @@ class eZOEInputParser extends eZXMLInputParser
                     if ( isset( $this->Namespaces[$prefix] ) )
                     {
                         $URI = $this->Namespaces[$prefix];
-                        $element->setAttributeNS( $URI, $qualifiedName, $value );
+                        $element->setAttributeNS( $URI, $qualifiedName, htmlspecialchars_decode( $value ) );
                     }
                     else
                     {
@@ -271,7 +271,7 @@ class eZOEInputParser extends eZXMLInputParser
                 }
                 else
                 {
-                    $element->setAttribute( $qualifiedName, $value );
+                    $element->setAttribute( $qualifiedName, htmlspecialchars_decode( $value ) );
                 }
             }
         }
--- a/extension/ezoe/ezxmltext/handlers/input/ezoexmlinput.php
+++ b/extension/ezoe/ezxmltext/handlers/input/ezoexmlinput.php
@@ -1744,12 +1744,12 @@ class eZOEXMLInput extends eZXMLInputHandler
                 if ( $customAttributePart === '' )
                 {
                     $customAttributePart = ' customattributes="';
-                    $customAttributePart .= $attribute->name . '|' . $attribute->value;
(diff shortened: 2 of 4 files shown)


brookinsconsulting:
+1

@pkamps  Excellent documentation. Do we have an issue ticket for this PR?

Cheers,
Brookins Consulting

peterkeung:
Issue has been created: https://jira.ez.no/browse/EZP-27171

andrerom:
@kmadejski Do you send this to QA?

kmadejski:
@andrerom JIRA issue status changed, but I can not change label here on Github.

adamwojs:
I see in JIRA that this is already approved by QA: https://jira.ez.no/browse/EZP-27171?focusedCommentId=211187&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-211187

PR #1290, opened 2017-03-29
EZP-27257: Handle non-numeric image size keys better

harmstyler:
Currently the fallback when looking for icon sizes grabs the icon with the key `0`, but there are instances where there is nothing set at that key. I assume the author was meaning to grab the first icon size in the array, so I propose switching this to use reset.

### DIFF
--- a/kernel/common/ezwordtoimageoperator.php
+++ b/kernel/common/ezwordtoimageoperator.php
@@ -268,7 +268,7 @@ class eZWordToImageOperator
                 }
                 else
                 {
-                    $size = $sizes[0];
+                    $size = reset($sizes);
                 }
 
                 $pathDivider = strpos( $size, ';' );


brookinsconsulting:
+1

yannickroger:
Thanks for the pull request. You'll need to create a JIRA issue (https://jira.ez.no) for this bug with steps to reproduce the problem so our quality insurance team can test it after the code review and before the merge.

pkamps:
An issue or steps to reproduce would be good to have. I think the array can be empty if the INI settings are not correctly configured. In that case, we should consider to check is the array is empty and in that case report an error and set $operatorValue to an empty string.

harmstyler:
Jira issue created: https://jira.ez.no/browse/EZP-27257

andrerom:
+1, and can be backported to 5.4

PR #1292, opened 2017-04-25
https://jira.ez.no/browse/EZP-27298: search index not updated if main…

bernhard-roessler:
Ticket: https://jira.ez.no/browse/EZP-27298
When moving the main location of a node, the object was not scheduled for reindexing - thus the index contained the old location of the object.


pkamps:
I'm not able to reproduce the issue. I tried:
* Created a node with 2 locations
* Moved the main node to a new location

My tests never reach the `else` part. It always has a `$nodeAssignment` and therefore calls `eZSearch::removeNodeAssignment` and `eZSearch::addNodeAssignment`

PR #1293, opened 2017-04-25
https://jira.ez.no/browse/EZP-27299: search index not updated if obje…

bernhard-roessler:
Ticket: https://jira.ez.no/browse/EZP-27299
Search index is not updated if a object is hidden by the cronjob (hide.php).

### DIFF
--- a/cronjobs/hide.php
+++ b/cronjobs/hide.php
@@ -49,6 +49,7 @@ foreach ( $rootNodeIDList as $nodeID )
                 {
                     $cli->output( 'Hiding node: "' . $node->attribute( 'name' ) . '" (' . $node->attribute( 'node_id' ) . ')' );
                     eZContentObjectTreeNode::hideSubTree( $node );
+					eZSearch::updateNodeVisibility( $node->attribute( 'node_id' ), 'hide' );
                 }
                 // clear memory after every batch
                 eZContentObject::clearCache();


gggeek:
Any reason why the standard hide operation would not work in this scenario? Maybe there is a bug somewhere to be fixed  instead of adding this workaround ?

siegfried-zach:
What do you mean with "standard hide operation"?
I assume the "eZContentObjectTreeNode::hideSubTree( $node );" is already a standard operation. In our cases it did not register the object for reindexing. Maybe there is a better place to put this change, but for us it was the "nearest" one :-)

c-revillo:
maybe @gggeek is talking about this one? https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/classes/ezcontentobjecttreenode.php#L5983
Looking there https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/classes/ezcontentobjecttreenode.php#L6020 i don't know if that op does a reindex of the content, but if not, maybe the code you added can be moved here...

andrerom:
And this is the problem with (missing) layers in legacy.. For the most part use of eZSearch is done outside of kernel/classes, like for instance in https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/content/ezcontentoperationcollection.php#L1011

pkamps:
I agree. The search index update should happen in https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/content/ezcontentoperationcollection.php.

Bernhard, would the `function changeHideStatus` work for you?

PR #1294, opened 2017-04-25
EZP-27300: unhide object if publish_date reached

bernhard-roessler:
This is a suggestion for Ticket: https://jira.ez.no/browse/EZP-27300


pkamps:
I think that feature should be wrapped in an extension and not added directly to the kernel.

Also, are you aware of https://github.com/mugoweb/hideuntildate. I think it's doing something very similar.

PR #1295, opened 2017-04-27
Fix EZP-27254: "Unknown relation type 0." error when access content o…

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-27254

# Steps to reproduce

> 1. Create an article;
> 2. Add an existing image to an XML block field. Used "Summary";
> 3. Add a different image to another XML block field. Used "Body";
> 4. Publish;
> 5. Edit the article, and publish;
> 6. Access the REST API via cURL: 
> curl -u "admin:admin" -v -H "Accept: application/vnd.ez.api.Content+json" http://example.com/api/ezp/v2/content/objects/136

# Problem description 

Problem is diagnosed in content classes with two or more XML block attributes. After second object update in ezcontentobject_link table there is created additional row with relation_type column equal 0 which causes "Unknown relation type 0." exception when object is accessed via API. 

# Backporting

~~~Patch can be also applied to 5.3 and probably in older versions of ezpublish-legacy (additional tests are required).~~~

Patch can be applied to all maintained releases. Originally the relationType condition was introduced in 71cd745ba0e71ea73683edc41bd20f69ce041c7b (line no. 2371) and wasn't modified until now. 

# TODO
- [x] ~~~Tests~~~
- [x] Backporting

### DIFF
--- a/kernel/classes/ezcontentobject.php
+++ b/kernel/classes/ezcontentobject.php
@@ -2873,7 +2873,7 @@ class eZContentObject extends eZPersistentObject
                   WHERE  from_contentobject_id=$fromObjectID AND
                          from_contentobject_version=$fromObjectVersion AND
                          to_contentobject_id=$toObjectID AND
-                         $relationTypeMatch != 0 AND
+                         $relationTypeMatch >= 0 AND
                          contentclassattribute_id=$attributeID";
         $count = $db->arrayQuery( $query );
         // if current relation does not exist


adamwojs:
review ping @andrerom @glye

glye:
> After second object update in ezcontentobject_link table there is created additional row with relation_type column equal 0

Why? Is that row supposed to be like that? I see how the fix would work, just want to make sure it's really a fix and not a workaround.

adamwojs:
Let's start from the beginning with an example content class `article` that has two `ezxmltexttype` fields named `editor_foo` and `editor_bar`.

1. User follows the first 4 steps: create `article` , insert the `kermit_the_frog.jpg` (T_CO_ID = 123) image into `editor_foo` field, the `fozzie_bear.jpg` (T_CO_ID = 126) image into `editor_bar` and publishes the content. At this point in `ezcontentobject_link` table there are 2 rows related to the newly created `article`:

| F_CO_ID | F_CO_V | ID  | RELATION_TYPE | T_CO_ID |
|---------|--------|-----|---------------|---------|
| 155        | 1          | 186 | 2             | 123     |
| 155        | 1          | 188 | 2             | 126     |

2. When user starts editing `article` the new draft is created. At this point the `ezcontentobject_link` table containts 2 new rows related to second version of `article`:

| F_CO_ID | F_CO_V | ID  | RELATION_TYPE | T_CO_ID |
|---------|--------|-----|---------------|---------|
| 155        | 1          | 186  | 2                      | 123     |
| 155        | 1          | 188  | 2                      | 126     |
| 155        | 2          | 189  | 2                      | 123     |
| 155        | 2          | 190  | 2                      | 126     |

3. The user publishes a draft. For each `ezxmltexttype` field:

3a. Input relations are collected with an `$object->appendInputRelationList(...)` call for each relation type and then committed by invoking `$object->commitInputRelations(...)`. See `kernel/classes/datatypes/ezxmltext/ezxmltexttype.php` lines 194-215:

```php
// linked objects
$linkedObjectIdArray = $this->getRelatedObjectList( $dom->getElementsByTagName( 'link' ) );

// embedded objects
$embeddedObjectIdArray = array_merge(
    $this->getRelatedObjectList( $dom->getElementsByTagName( 'embed' ) ),
    $this->getRelatedObjectList( $dom->getElementsByTagName( 'embed-inline' ) )
);

if ( !empty( $embeddedObjectIdArray ) )
{
    $object->appendInputRelationList( $embeddedObjectIdArray, eZContentObject::RELATION_EMBED );
}

if ( !empty( $linkedObjectIdArray ) )
{
    $object->appendInputRelationList( $linkedObjectIdArray, eZContentObject::RELATION_LINK );
}
if ( !empty( $linkedObjectIdArray ) || !empty( $embeddedObjectIdArray ) )
{
    $object->commitInputRelations( $currentVersion->attribute( 'version' ) );
}
```
</details>

The input relation list is common for all processed fields (relations are aggregated here).

3b.  `$object->commitInputRelations(...)` computes diff between input relation list and the already persisted relations of content object. See `kernel/classes/ezcontentobject.php` lines 2329-2351:

```php
foreach ( $this->InputRelationList as $relationType => $relatedObjectIDArray )
{
    $oldRelatedObjectArray = $this->relatedObjects( $editVersion, false, 0, false, array( 'AllRelations' => $relationType ) );

    foreach ( $oldRelatedObjectArray as $oldRelatedObject )
    {
        $oldRelatedObjectID = $oldRelatedObject->ID;
        if ( !in_array( $oldRelatedObjectID, $relatedObjectIDArray ) )
        {
            $this->removeContentObjectRelation( $oldRelatedObjectID, $editVersion, 0, $relationType );
        }
        $relatedObjectIDArray = array_diff( $relatedObjectIDArray, array( $oldRelatedObjectID ) );
    }

    foreach ( $relatedObjectIDArray as $relatedObjectID )
    {
        $this->addContentObjectRelation( $relatedObjectID, $editVersion, 0, $relationType );
    }
}
```  

Each record in the `ezcontentobject_link` table represents a relation between two objects with types specified in the `relation_type` column. So changes are persisted by switching particular bit flags.

Let's go back to our example: After relations collected from `editor_foo` field are committed, the `ezcontentobject_link` table look like this:  

| F_CO_ID | F_CO_V | ID  | RELATION_TYPE | T_CO_ID |
|---------|--------|-----|---------------|---------|
| 155        | 1          | 186  | 2                      | 123        |
| 155        | 1          | 188  | 2                      | 126        |
| 155        | 2          | 189  | 2                      | 123       |
| 155        | 2          | 190  | **0**               | 126       |

Row with id = 190 contains 0 because at this point relation `fozzie_bear.jpg` was not found on the input relation list.  

Then `editor_bar` is processed and relation to `fozzie_bear.jpg` is added to the input relation list, which contains 2 items now: `kermit_the_frog.jpg` and `fozzie_bear.jpg`. Relation to  `kermit_the_frog.jpg` is updated as usual, but system couldn't find the row for `fozzie_bear.jpg` (it is searching for rows with **relation_type != 0**), so that's why we get an additional row.  

The final state of `ezcontentobject_link` table:

| F_CO_ID | F_CO_V | ID  | RELATION_TYPE | T_CO_ID |
|---------|--------|-----|---------------|---------|
| 155        | 1          | 186  | 2                      | 123        |
| 155        | 1          | 188  | 2                      | 126        |
| 155        | 2          | 189  | 2                      | 123       |
| 155        | 2          | 190  | **0**               | 126       |
| 155        | 2          | 186  | 2                      | 123        |
| 155        | 2          | 188  | 2                      | 126        |

pkamps:
I did not test this pull request but I trust that it is fixing those extra DB rows with a relation_type = 0.

I think those rows still get created in the DB but then get updated with a proper relation_type (_eZContentObject::addContentObjectRelation_).

I checked if we can avoid those rows in the first place - then we do not need to change the SQL statement in _eZContentObject::addContentObjectRelation_. You changed the condition to "$relationTypeMatch >= 0" which is basically removing the condition.

Let me explain where I think the problem is:
Function _eZXMLTextType::onPublish_ is called for each ezxmltext field if you publish a new version.
In the example above, it gets called twice because there are the two ezxmltext fields called "Summary" and "Body".

Function _onPublish_ collects all relations for one given field. So it searches the "Summary" ezxml text for embedded objects or linked objects. With that list of object IDs, it is calling
_eZContentObject::commitInputRelations_. That function compares the given list of object IDs (```$this->InputRelationList```) with all object relations that object currently has.
And that's the problem: all object relations of the current object contains relations from other attributes as well (for example the "Body" field could have embedded objects as well).

_eZContentObject::commitInputRelations_ works in context of an object. But if you use that function in context of a single field like _eZXMLTextType::onPublish does, _eZContentObject::commitInputRelations_ will start to remove and add object relations that you don't really need to touch.

You cannot change the function _eZContentObject::commitInputRelations_ to only compare object relations for a single field (attribute): the relation types 'embed' and 'link' do not store the _contentclassattribute_id_ in the _ezcontentobject_link_ DB table.

So you are forced to always compare all embed and link relations for the entire object and not in context of a single field/attribute.

I think a better approach to fix this bug is to build a list of all embed/linked object relations in _eZXMLTextType::onPublish_  for all ezxmltext fields on the object. And send that list to _eZContentObject::commitInputRelations_.
It will also avoid those extra DB create/delete statements while comparing the object relations lists.

PR #1297, opened 2017-05-05
Fix EZP-27321: Update TinyMCE to 3.5.12 and patch for "Uncaught DOMEx…

adamwojs:
…ception: Failed to execute 'setBaseAndExtent' on 'Selection': There is no child at offset 1."

> JIRA: https://jira.ez.no/browse/EZP-27321

# Description

This PR updates TinyMCE version from **3.5.10** to **3.5.12** ([changelog](https://github.com/adamwojs/tinymce/blob/fix_failed_to_execute_setBaseAndExtent_on_selection_there_is_no_child_at_offset_1/changelog.txt)) with **additional fix** of following error `Uncaught  DOMException: Failed to execute 'setBaseAndExtent' on 'Selection': There is no child at offset 1`. (https://github.com/adamwojs/tinymce/commit/7a2c6fbdb9066fc6f51cfbdbdd7a802f79154a92)

# Steps to reproduce

This bug was diagnosed only in **Chrome 58**.  

1. Create a new article
2. Add an embed image to an XML block
3. Click on the image 

The error above will be returned in **browser console**.


andrerom:
great work so far.

Now for the hard part, if my memory serves me right we already had a few patches applied to TinyMCE, some not from upstream, so if you did a hard owerwrite here they are probably wiped out.

Problem is I don't remember the details anymore, might be this rings a bell for @dpobel, but we might need to dig into the revision history to uncover what we have applied over time.

adamwojs:
Unfortunately this patch brings TinyMCE build from upstream repository fork (3.X branch), so any changes in our repository are overwritten.  

I will try to figure out what is difference between original TinyMCE repository and our version, and prepare build with all overwritten patches.

andrerom:
Actually , assuming there are not changes to TinyMCE itself since b09f7086b96551cfa5c71742cd2293b7cb45d420 I guess we are good on that. Looking a bit more around if I remember right some features was forked to avoid exactly what I was afraid of above. This goes for:
- plugins/{eztable, ezemotions}
- themes/ez _(fork of advance)_

And in `ezoe/` you have custom stuff.

So remaining work here is make sure to manually apply updates to those plugins and the theme from it's original version that you have now updated, from what I can see both advance theme and table plugin got some fixes in this update.

extension/ezoe/design/standard/javascript/plugins/eztable/editor_plugin.js
```diff
@@ -971,33 +1015,36 @@
...
 									return;
 								}
 
 								// BR element
 								if (node.nodeName == 'BR') {
-									if (start)
-										rng.setStartBefore(node);
-									else
-										rng.setEndBefore(node);
+                                    if (start) {
```
andrerom: whitespace issues here? :)

adamwojs: Yes, my fault. I need to merge last one patch (https://github.com/adamwojs/tinymce/commit/b85c2cda8d7b44235f1b51bb907e543b7f667b79) so I'm still working on this PR :-)

adamwojs:
PR Update:

* `ezemotions ` plugin doesn't require any changes
* Most of the changes in `eztable` plugin are CS fixes (missing open/close braces, missing radix argument in `parseInt` invocation etc.)    
* I've skipped merge of https://github.com/adamwojs/tinymce/commit/b85c2cda8d7b44235f1b51bb907e543b7f667b79 patch because in `ez` theme. There are too many changes in our fork, to apply these changes in another way than reimplement this feature.

andrerom:
> I skipped merge of adamwojs/tinymce@b85c2cd patch in ez theme. There are too many changes in our fork, to apply these changes in another way than reimplement this feature.

👍  No need to add this feature in our theme at this point

PR #1299, opened 2017-06-07
EZP-27714: Replaced wrong variable name in eZTemplateMultiPassParser class

amauryvallier:
Using $tagPosition brings a notice that it does not exists => the variable to use is $tagPos


amauryvallier:
I created the issue (with the right informations I hope) : https://jira.ez.no/browse/EZP-27714

PR #1300, opened 2017-06-16
EZP-27460 proper solr indexing while moving content

tomaszmadeyski:
This is a fix for https://jira.ez.no/browse/EZP-27460

## Description
While moving content in eZ there was no action taken to update information in search engine about children of node being moved.

This PR adds a method which recursively updates information in search about given node.

### DIFF
--- a/kernel/classes/ezcontentobjecttreenodeoperations.php
+++ b/kernel/classes/ezcontentobjecttreenodeoperations.php
@@ -113,6 +113,7 @@ class eZContentObjectTreeNodeOperations
                 $nodeIDList = array( $nodeID );
                 eZSearch::removeNodeAssignment( $node->attribute( 'main_node_id' ), $newNode->attribute( 'main_node_id' ), $object->attribute( 'id' ), $nodeIDList );
                 eZSearch::addNodeAssignment( $newNode->attribute( 'main_node_id' ), $object->attribute( 'id' ), $nodeIDList, true );
+                eZSearch::refreshChildren( $node );
             }
 
             $result = true;
--- a/kernel/classes/ezsearch.php
+++ b/kernel/classes/ezsearch.php
@@ -671,7 +671,7 @@ class eZSearch
         {
             foreach ( $node->children() as $child )
             {
-                self::addNodeAssignment( $child->MainNodeID, $child->ContentObjectID, array( $child->MainNodeID ), true);
+                self::addNodeAssignment( $child->MainNodeID, $child->ContentObjectID, array( $child->MainNodeID ), true );
                 self::refreshChildren( $child );
             }
        }
--- a/kernel/classes/ezsearch.php
+++ b/kernel/classes/ezsearch.php
@@ -671,7 +671,7 @@ class eZSearch
         {
             foreach ( $node->children() as $child )
             {
-                self::addObject( $child->ContentObject );
+                self::addNodeAssignment( $child->MainNodeID, $child->ContentObjectID, array( $child->MainNodeID ), true);
                 self::refreshChildren( $child );
             }
        }
--- a/kernel/classes/ezsearch.php
+++ b/kernel/classes/ezsearch.php
@@ -659,6 +659,23 @@ class eZSearch
 
         return false;
     }
+
+    /**
+     * Notifies search engine to refresh information about children of given node
+     *
+     * @param eZContentObjectTreeNode $node
+     */
+    public static function refreshChildren( $node )
+    {
+        if ( $node->childrenCount( false ) > 0 )
+        {
+            foreach ( $node->children() as $child )
+            {
+                self::addObject( $child->ContentObject );
+                self::refreshChildren( $child );
+            }
+       }
+    }
 }
 
 ?>


pkamps:
I see memory/timeout issues moving bigger subtrees.
I think you can only solve that issue with delayed indexing or a threshold to limit the children you re-index.

andrerom:
@pkamps could be, but from @tomaszmadeyski's digging most other operations on subtree uses same approach. or?

tomaszmadeyski:
Indeed. This approach is inspired by changing section of an object which changes section of whole subtree. 

So [here](https://github.com/ezsystems/ezfind/blob/master/search/plugins/ezsolr/ezsolr.php#L1431) is the method for updating section in ezfind. It takes **an array** of ids and calls `n` times [registerSearchObject](https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/content/ezcontentoperationcollection.php#L615) which eventually calls [addObject](https://github.com/ezsystems/ezfind/blob/master/search/plugins/ezsolr/ezsolr.php#L448) (this is what I'm doing in this PR). It is called from [here](https://github.com/ezsystems/ezpublish-legacy-ee/blob/5.4/kernel/classes/ezcontentobjecttreenode.php#L2862) - there's array of `nodeIds` built explicitly

glye:
Any recursive subtree operation can die this way, and any added processing increases the risk - but that said, what is the risk in this particular case? That is, how heavy is the indexing, compared to existing recursive operations? It does need a lot of reads.

tomaszmadeyski:
@glye Indexing itself is exact the same operation in case of changing section of subtree and moving subtree: search index is rebuilt for all objects under the root node of subtree. 

In general moving subtree is lighter than updating section of subtree as in the section update each node is [fetched from database](https://github.com/ezsystems/ezfind/blob/master/search/plugins/ezsolr/ezsolr.php#L1435) and while moving we have all objects already in place (AFAIK).

pkamps:
I think the biggest risk is the memory limit. The move operation will fetch all objects (and store in memory) for a subtree. I'm sure you will hit the PHP memory limit for bigger subtrees. The editorial experience is bad -- it will show a fatal error page.

It is not just a theoretical problem: In eZContentObjectTreeNodeOperations::move you can see that cache manager is using a threshold (https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/classes/ezcontentobjecttreenodeoperations.php#L129)
I believe that's for the same reason - clearing cache for bigger subtrees is problematic.

I don't think we should justify this pull request because other tree operations are doing the same (like updating the section ID): The operation of updating the section ID has the same memory/timeout issues.

OK - so much for the nagging part. We want to update search index on move operations and it is actually solved if you use "Delayed indexing". It is much more harder if you don't want to delay the indexing and have it done on page load. I actually don't have an idea how to solve it.
We could use a threshold limit and just not re-index subtrees if the child count exceeds the threshold (and write a log entry).
There is no "re-index all" for search indexes - so we cannot use the same approach as the CacheManager.

andrerom:
@pkamps For the constructive part, could you point to example of delayed indexing perhaps so @tomaszmadeyski can have something to consider?

peterkeung:
Here's the delayed indexing script which handles moved nodes: https://github.com/ezsystems/ezpublish-legacy/blob/master/cronjobs/indexcontent.php

Here's the ticket where that "index move nodes" feature was added in 2014: http://jira.ez.no/browse/EZP-23681

tomaszmadeyski:
@peterkeung, @pkamps  this looks good! I will try to make use of this delayed indexing. Thanks for the input

tomaszmadeyski:
Ok guys, I commited a change in 55a0272 
Basically I'm making use of `addNodeAssignment` method instead of `addObject`. It calls [addNodeAssignment](https://github.com/ezsystems/ezfind/blob/master/search/plugins/ezsolr/ezsolr.php#L1494) in ezsolr which calls [registerSearchObject](https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/content/ezcontentoperationcollection.php#L571) which has support for delayed indexing

kernel/classes/ezcontentobjecttreenodeoperations.php
```diff
@@ -113,6 +113,7 @@ static function move( $nodeID, $newParentNodeID )
                 $nodeIDList = array( $nodeID );
                 eZSearch::removeNodeAssignment( $node->attribute( 'main_node_id' ), $newNode->attribute( 'main_node_id' ), $object->attribute( 'id' ), $nodeIDList );
                 eZSearch::addNodeAssignment( $newNode->attribute( 'main_node_id' ), $object->attribute( 'id' ), $nodeIDList, true );
+                eZSearch::refreshChildren( $node );
```
glye: Is this still needed? I thought the delayed indexer handled subtrees by itself? Ref https://github.com/ezsystems/ezpublish-legacy/blob/master/cronjobs/indexcontent.php#L56

glye: Hmmmnope, if that was the case, the whole PR would be redundant. But then I'm not sure what is meant by code comment linked above.

andrerom: same comment here, from how I understood it, and based on the call one line above here, we basically need to make sure customer uses this indexing cronjob, OR maybe rather we need to make sure [ezplatformsearch](https://github.com/netgen/ezplatformsearch) handles moved node actions.

Either way we need to update doc somewhere, probably in ezplatformsearch extension.

tomaszmadeyski: ah, right, my bad, I didn't check what is actually happening in cronjob task :(
Once all is configured well, meaning `DelayedIndexing` is enabled and cronjob is setup there should be no issue and this is not needed anymore. 

But from the other hand, assuming that `DelayedIndexing` is disabled (it is disabled by default), without adding this `refreshChildren` call there will be no reindex of children at all and this looks like an issue (it was reported after all, that's where this PR comes from). 

There are probably few ways to fix it:
1. "Force" users to use delayed indexing / inform them that there will be no reindex if they don't
2. Keep `refreshChildren` method but in this case we need to do something with the way it is called as currently it adds task to reindex every single node of subtree and cronjob takes each of such node and reindexes whole subtree so there's a lot of redundancy here as single node can be reindexed multiple times

glye: Check `DelayedIndexing` and only run the recursive indexer if it is disabled?

tomaszmadeyski: yes, this seems legit. @andrerom what do you think?

kernel/classes/ezsearch.php
```diff
@@ -659,6 +659,23 @@ public static function swapNode( $nodeID, $selectedNodeID, $nodeIdList = array()
...
     }
+
+    /**
+     * Notifies search engine to refresh information about children of given node
+     *
+     * @param eZContentObjectTreeNode $node
+     */
+    public static function refreshChildren( $node )
+    {
+        if ( $node->childrenCount( false ) > 0 )
```
bdunogier:
```
if ( $node->childrenCount( false ) === 0 ) {
  return;
}
```

Then you can remove one nesting level.

kernel/classes/ezcontentobjecttreenodeoperations.php
andrerom: > yes, this seems legit. @andrerom what do you think?

If there are examples already doing this, maybe.
But I'd rather prefer we force DelayedIndexing from LegacyBridge / LegacyBundle if we know we depend on it, but then either doc will need to cover that this needs to be run, OR as hinted above  ezplatformsearch needs to take this action into account here in it's included cronjob: 

https://github.com/netgen/ezplatformsearch/blob/master/cronjobs/ezplatformindexsubtree.php

@emodric Wdyt?

peterkeung: If we follow what was implemented for delayed indexing, the logic should be put in eZContentOperationCollection::registerSearchObject(). If we follow eZSearch::addNodeAssignment, the 4th parameter for that is $isMoved, which then gets propagated to the search engine implementation and eZContentOperationCollection::registerSearchObject().

Because of the risk of timeouts, I think we should simply make a cronjob a requirement, regardless of whether the delayed indexing flag is on or not. As it is now, the delayed indexing flag is a requirement: https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/content/ezcontentoperationcollection.php#L571.

emodric: Using `addNodeAssignment` when we really want to reindex a subtree feels wrong to me. I agree with @peterkeung and @pkamps here. Require cronjob to be run, that is, act as if delayed indexing is turned on.

I have no problems with adding a case for moved nodes to `ezplatformsearch` and documenting it.

andrerom: > I have no problems with adding a case for moved nodes to ezplatformsearch and documenting it.

That would be great, and about the right place to solve it for solrBundle case right?
Should probably also detect if delayed indexing is disabled and log warning/error in all ezplatformsearch methods that call "registerSearchObject" btw.

andrerom: Or would ezplatformsearch work fine with https://github.com/ezsystems/ezpublish-legacy/blob/master/cronjobs/indexcontent.php ? Which handles the kernel pending index actions? I'm also fine with that, but then we need to expand doc in ezplatformsearch for that right?

emodric: @andrerom I think that both scripts need to be active. The from legacy to handle standard delayed indexing actions, and the from eZ Platform Search, to handle changing the visibility of a node.

Look here: https://github.com/netgen/ezplatformsearch/pull/3

Ping @harmstyler 

I would like to see them merged into one somehow, however.

andrerom: @emodric / @peterkeung One issue found with just using delayed indexing as is, http cache will get out of date (purged before index is updated). We have similar issues when soft commit is set to a high number, but in this case it will be worse as it will be both soft commit time and interval of the cronjob which would typically be much longer that the soft commit time.

We might need to emit an indexing event here on index so new stack can purge the affected locations.

pkamps:
That reminds me this pull request: https://github.com/ezsystems/ezpublish-legacy/pull/1210

andrerom:
> That reminds me this pull request: #1210

True, that would solve it without having to introduce new events.

@tomaszmadeyski would that PR solve the remaining question?

tomaszmadeyski:
> would that PR solve the remaining question?

I believe so if only this clears http cache as well

andrerom:
> I believe so if only this clears http cache as well

it should do: https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/classes/ezcontentcachemanager.php

It will trigger `content/cache` or `content/cache/all` which is listened to here: https://github.com/ezsystems/LegacyBridge/blob/master/bundle/LegacyMapper/Configuration.php#L207

It will also clear persistence cache which is a bit waste, but it's ok for legacy bridge, internally in new stack that would not have been desirable.

pkamps:
I think you can use an event when object cache is cleared. That's what other "varnish purge on publish" solutions do:
https://github.com/ezsystems/ezpublish-legacy/pull/119

I hope that event (static cache handler) gets triggered, I did not verify it.

andrerom:
> I hope that event (static cache handler) gets triggered, I did not verify it.

I guess it should. In legacyBundle/Bridge case it listens to ezpEvent.

andrerom:
Closing as #1210 has been merged for this.

PR #1301, opened 2017-06-16
EZP-25916: fixing deadlock issues on clustering cache

tomaszmadeyski:
This is a fix for https://jira.ez.no/browse/EZP-25916 

## Description
In high load environments it is possible that multiple processes are trying to perform the same operation (cache generation) on mysql database which may result in one of [Lock wait timeout exceeded; try restarting transaction](https://dev.mysql.com/doc/refman/5.5/en/error-messages-server.html#error_er_lock_wait_timeout) (error number 1205) or [Deadlock found when trying to get lock; try restarting transaction](https://dev.mysql.com/doc/refman/5.5/en/error-messages-server.html#error_er_lock_deadlock) (error number 1213)

Currently those errors aren't handled and are treated as a `Unexpected error` resulting with a crash.

The idea behind this solution is to treat those errors similar as [1062](https://dev.mysql.com/doc/refman/5.5/en/error-messages-server.html#error_er_dup_entry) error is treated already, by returning array result which will force DFS cache to read from stale cache file.

### DIFF
--- a/kernel/private/classes/clusterfilehandlers/dfsbackends/mysqli.php
+++ b/kernel/private/classes/clusterfilehandlers/dfsbackends/mysqli.php
@@ -1638,8 +1638,11 @@ class eZDFSFileHandlerMySQLiBackend implements eZClusterEventNotifier
         if ( !$this->_query( $query, "_startCacheGeneration( $filePath )", false ) )
         {
             $errno = mysqli_errno( $this->db );
+
             //once deadlock found / lock wait timeout found, we return ko status which will force read from stale cache
             //'remaining' item is irrelevant in this case, but it needs to be set
+            //1205 is 'Lock wait timeout exceeded; try restarting transaction'
+            //1213 is 'Deadlock found when trying to get lock; try restarting transaction'
             if ( in_array( $errno, array( 1205, 1213 ) ) )
             {
                 return array( 'result' => 'ko', 'remaining' => 0 );
--- a/kernel/private/classes/clusterfilehandlers/dfsbackends/mysqli.php
+++ b/kernel/private/classes/clusterfilehandlers/dfsbackends/mysqli.php
@@ -1653,7 +1653,7 @@ class eZDFSFileHandlerMySQLiBackend implements eZClusterEventNotifier
                 eZDebug::writeError( "Unexpected error #$errno when trying to start cache generation on $filePath (".mysqli_error( $this->db ).")", __METHOD__ );
                 eZDebug::writeDebug( $query, '$query' );
 
-                // @todo Make this an actual error, maybe an exception
+                // @todo throw an exception
                 return array( 'result' => 'error' );
             }
             // error 1062 is expected, since it means duplicate key (file is being generated)
--- a/kernel/private/classes/clusterfilehandlers/dfsbackends/mysqli.php
+++ b/kernel/private/classes/clusterfilehandlers/dfsbackends/mysqli.php
@@ -1654,7 +1654,7 @@ class eZDFSFileHandlerMySQLiBackend implements eZClusterEventNotifier
                 eZDebug::writeDebug( $query, '$query' );
 
                 // @todo Make this an actual error, maybe an exception
-                return array( 'res' => 'ko' );
+                return array( 'result' => 'error' );
             }
             // error 1062 is expected, since it means duplicate key (file is being generated)
             else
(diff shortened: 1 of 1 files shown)


kernel/private/classes/clusterfilehandlers/dfsbackends/mysqli.php
```diff
@@ -1638,6 +1638,13 @@ public function _startCacheGeneration( $filePath, $generatingFilePath )
         if ( !$this->_query( $query, "_startCacheGeneration( $filePath )", false ) )
         {
             $errno = mysqli_errno( $this->db );
+            //once deadlock found / lock wait timeout found, we return ko status which will force read from stale cache
+            //'remaining' item is irrelevant in this case, but it needs to be set
+            if ( in_array( $errno, array( 1205, 1213 ) ) )
```
andrerom: maybe expand on what 1205 and 1213 is, like done for 1062 below

kernel/private/classes/clusterfilehandlers/dfsbackends/mysqli.php
```diff
@@ -1638,6 +1638,13 @@ public function _startCacheGeneration( $filePath, $generatingFilePath )
         if ( !$this->_query( $query, "_startCacheGeneration( $filePath )", false ) )
         {
             $errno = mysqli_errno( $this->db );
+            //once deadlock found / lock wait timeout found, we return ko status which will force read from stale cache
```
andrerom: newline

kernel/private/classes/clusterfilehandlers/dfsbackends/mysqli.php
```diff
@@ -1638,6 +1638,16 @@ public function _startCacheGeneration( $filePath, $generatingFilePath )
         if ( !$this->_query( $query, "_startCacheGeneration( $filePath )", false ) )
         {
             $errno = mysqli_errno( $this->db );
+
+            //once deadlock found / lock wait timeout found, we return ko status which will force read from stale cache
```
andrerom: super nitpick, but as seen on line further down we always have a whitespace before the comment text

gggeek:
Going a bit OT hewre, but a couple of lines below your change there is a line that reads
`return array( 'res' => 'ko' );`

If we look at the only place in the codebase which calls startCacheGeneration(), we see that having a 'ret' return member is not supported, whereas we would expect the 'return' member to be set.

What about changing that line to
`return array( 'result' => 'error' );`
?

tomaszmadeyski:
@gggeek yes, you are right. My change was made with "this is legacy code, we should fix bugs, no refactoring included" in mind. @andrerom is up to you to decide, should I also fix this small bug? (it is about [this line](https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/private/classes/clusterfilehandlers/dfsbackends/mysqli.php#L1647))

andrerom:
If `res` is not handled, then yes please propose that in separate commit here, and we'll run it by BD afterwards to make sure we haven't missed anything.

tomaszmadeyski:
@gggeek done in b64dea3

tomaszmadeyski:
> Did you find ways to test this fix

Yes, I did stuff from Jira task description. Basically to reproduce deadlock issue in mysql the easiest way is to open mysql console, start transaction, perform the same insert as eZ Publish is going to do and does not commit it.

```
MariaDB [foobar]> START TRANSACTION;
Query OK, 0 rows affected (0.00 sec)
 
MariaDB [foobar]> INSERT INTO ezdfsfile_cache ( name, name_trunk, name_hash, scope, datatype, mtime, expired ) VALUES('var/cache/user-info/1/0/user-data-10.cache.php.generating', 'var/cache/user-info/1/0/user-data-10.cache.php.generating', MD5('var/cache/user-info/1/0/user-data-10.cache.php.generating'), '', '', 1466062010, 0);
Query OK, 1 row affected (0.00 sec)
 
MariaDB [foobar]>
```

In fact, only value which is important here is `name_hash` as this is a primary key of `ezdfsfile_cache` table

kernel/private/classes/clusterfilehandlers/dfsbackends/mysqli.php
```diff
@@ -1638,13 +1638,23 @@ public function _startCacheGeneration( $filePath, $generatingFilePath )
...
+            {
+                return array( 'result' => 'ko', 'remaining' => 0 );
+            }
+
             if ( $errno != 1062 )
             {
                 eZDebug::writeError( "Unexpected error #$errno when trying to start cache generation on $filePath (".mysqli_error( $this->db ).")", __METHOD__ );
                 eZDebug::writeDebug( $query, '$query' );
 
                 // @todo Make this an actual error, maybe an exception
```
glye: Does the line below qualify as an "actual error" now, so that this comment should be changed? :)

tomaszmadeyski: Changed in b67c4ca 
In general, I left a `todo` since a valid way to handle errors would be throwing an exception

PR #1302, opened 2017-06-16
Fix EZP-27474: Unable to add the same URL alias for other languages

natanael89:
[https://jira.ez.no/browse/EZP-27474](https://jira.ez.no/browse/EZP-27474)
#
Excerpt from the JIRA issue:
> When trying to add an URL alias to a node (content/urlalias/) where the same URL alias already exists in another language, the following error message is produced: The URL alias <</xxx>> already exists, and it points to <<content/view/full/xxx>>. As it is an alias for another language, it should be created anyway.

The solution proposed in this PR was originally proposed by @glye here: https://jira.ez.no/browse/EZP-16680. It was tested by me and it solves the issue.

### DIFF
--- a/kernel/classes/ezurlaliasml.php
+++ b/kernel/classes/ezurlaliasml.php
@@ -842,6 +842,15 @@ class eZURLAliasML extends eZPersistentObject
                         $curElementID = $curID;
                         break;
                     }
+
+                    // If the current node is the same action, but the language is different
+                    // (enables adding the same URL alias for other languages)
+                    if ( !( (int)$row['lang_mask'] & $languageID ) )
+                    {
+                        // We can reuse the element so record the ID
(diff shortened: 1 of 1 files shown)


bdunogier:
Inline comment is a good idea, as there is no way we can make a private method out of this :)

natanael89:
I have added the comments.

tharkun:
I love this PR. I manage multiple sites with a lot of similar languages (eng-GB, eng-CA, eng-US......) and we face this problem.

If it works, i'll upgrade quiclky

glye:
@natanael89 The reviews are good. It's ready to send to QA and then merge (hopefully), right?

natanael89:
Sure, I'm sending it to the QA.

natanael89:
Accepted by QA.

PR #1305, opened 2017-07-04
Fix EZP-27588: Warning shown when renaming files in concurrency situations

natanael89:
> [EZP-27588](https://jira.ez.no/browse/EZP-27588)
#
Excerpt from JIRA:
> The following warning is shown in rare situations on the customer's page: "Warning: unlink(var/log/error.log.1): No such file or directory in /var/www/xxx/ezpublish_legacy/lib/ezfile/classes/ezfile.php".
Unfortunately, I wasn't able to reproduce it on local installation but based on this information I think I know why it is happening and I have a good idea what can be done to fix it.
In my opinion, the warning is the result of the log rotation. Theoretically, if two rotations for the same file are run simultaneously, there could be a case that a file is already moved. If that's the case, there actually could be a warning thrown from this line: https://github.com/ezsystems/ezpublish-legacy/blob/master/lib/ezfile/classes/ezfile.php#L172. 
The easiest fix would be to add the error suppression operator "@" before the unlink function, just like it is done above (https://github.com/ezsystems/ezpublish-legacy/blob/master/lib/ezfile/classes/ezfile.php#L157).

This is the easiest fix mentioned above, meaning it suppresses warnings coming from the unlink function like it is done already in another place in the same function.

### DIFF
--- a/lib/ezfile/classes/ezfile.php
+++ b/lib/ezfile/classes/ezfile.php
@@ -169,7 +169,7 @@ class eZFile
         	    eZDebug::writeWarning( "$srcFile could not be renamed to $destFile", __METHOD__ );
 
             if ( $flags & self::CLEAN_ON_FAILURE )
-        	    unlink( $srcFile );
+        	    @unlink( $srcFile );
         }
 
         return $status;


dougplant:
I would recommend against using the "@" to suppress the warning. If a dev happens to have 'halt on warning' set, this type of situation causes a significant loss of time tracking down a silent crash.

I know this because I recently patched a similar case in lib/ezfile/classes/ezdir.php. In that case, it was a simple matter to make the operation more robust and make the error handling explicit.

natanael89:
Closing the PR for now as the issue no longer appears for the customer.

pkamps:
I have seen this error message in the past. So I tried to reproduce the issue: I wrote a script spinning off multiple processes and each writing to the log file. Unfortunately I was not able to get that error message.

gggeek:
I know about this warning and tbh I lived for years with it. Using @ to suppress any errors gets -1 from me as it might hide more serious stuff.

PR #1310, opened 2017-08-01
EZP-26013 moving commit before method which publishes content

tomaszmadeyski:
This is a fix for https://jira.ez.no/browse/EZP-26013 and follow up to #1306 

Previous PR was only partial fix. It worked only in case item restored from trash was an object of reverse relation. 
In order to fix it for all cases `commit` need to go further up, before method which is doing content publish.

//CC @lserwatka

### DIFF
--- a/kernel/content/restore.php
+++ b/kernel/content/restore.php
@@ -134,11 +134,7 @@ if ( $module->isCurrentAction( 'AddLocation' ) )
     $version->store();
 
     $object->restoreObjectAttributes();
-
     $user = eZUser::currentUser();
-
-    $db->commit();
-
     $operationResult = eZOperationHandler::execute( 'content', 'publish', array( 'object_id' => $objectID,
                                                                                  'version' => $version->attribute( 'version' ) ) );
     if ( ( array_key_exists( 'status', $operationResult ) && $operationResult['status'] != eZModuleOperationInfo::STATUS_CONTINUE ) )
@@ -166,7 +162,11 @@ if ( $module->isCurrentAction( 'AddLocation' ) )
         }
     }
 
-    eZContentObject::fixReverseRelations( $objectID, 'restore' );
+    eZContentObject::fixReverseRelations( $objectID, 'restore', false );
+    $db->commit();
+
+    // we need to clear cache again after db transcation is commited
+    eZContentCacheManager::clearContentCacheIfNeeded( $objectID,  $version->attribute( 'version' ) );
 
     $module->redirectToView( 'view', array( 'full', $mainNodeID ) );
     return;
--- a/kernel/content/restore.php
+++ b/kernel/content/restore.php
@@ -134,6 +134,7 @@ if ( $module->isCurrentAction( 'AddLocation' ) )
     $version->store();
 
     $object->restoreObjectAttributes();
+
     $user = eZUser::currentUser();
     $operationResult = eZOperationHandler::execute( 'content', 'publish', array( 'object_id' => $objectID,
                                                                                  'version' => $version->attribute( 'version' ) ) );
@@ -163,6 +164,7 @@ if ( $module->isCurrentAction( 'AddLocation' ) )
     }
 
     eZContentObject::fixReverseRelations( $objectID, 'restore', false );
+
     $db->commit();
 
     // we need to clear cache again after db transcation is commited
--- a/kernel/content/restore.php
+++ b/kernel/content/restore.php
@@ -136,6 +136,9 @@ if ( $module->isCurrentAction( 'AddLocation' ) )
     $object->restoreObjectAttributes();
 
     $user = eZUser::currentUser();
+
+    $db->commit();
+
     $operationResult = eZOperationHandler::execute( 'content', 'publish', array( 'object_id' => $objectID,
                                                                                  'version' => $version->attribute( 'version' ) ) );
     if ( ( array_key_exists( 'status', $operationResult ) && $operationResult['status'] != eZModuleOperationInfo::STATUS_CONTINUE ) )
@@ -162,8 +165,6 @@ if ( $module->isCurrentAction( 'AddLocation' ) )
             eZUser::purgeUserCacheByUserId( $object->attribute( 'id' ) );
         }
     }
-
-    $db->commit();
 
     eZContentObject::fixReverseRelations( $objectID, 'restore' );
 
--- a/kernel/classes/ezcontentobject.php
+++ b/kernel/classes/ezcontentobject.php
@@ -1762,8 +1762,9 @@ class eZContentObject extends eZPersistentObject
      *
      * @param int $objectID
      * @param string|bool $mode See eZObjectRelationListType::fixRelatedObjectItem() for valid modes
+     * @param bool $clearCacheIfEnabled If set to false cache won't be cleared
      */
-    static function fixReverseRelations( $objectID, $mode = false )
+    static function fixReverseRelations( $objectID, $mode = false, $clearCacheIfEnabled = true )
     {
         $db = eZDB::instance();
         $objectID = (int) $objectID;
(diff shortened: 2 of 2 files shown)


andrerom:
😕 .. Aka: Problem here is that it's very hard to know if this will have side effects, the transaction has spanned the whole function for 15 years, and now a few years after we released last 5.x release we are risking db inconsistencies to be able to fix edge cases for platform <> legacy integration which it was never meant to be perfect at in the first place. 

Anyway: Is content publish operation outside operation transaction other places?

tomaszmadeyski:
@andrerom 

> Is content publish operation outside operation transaction other places?

I can't tell with 100% certainty as I'm basing just on a static code analysis. I found usages of `eZOperationHandler::execute( 'content', 'publish' )` and I found several usages with **no transaction** at all:

1. [ezcsvimport.php](https://github.com/ezsystems/ezpublish-legacy/blob/master/bin/php/ezcsvimport.php#L127)
2. [ezsubtreecopy.php](https://github.com/ezsystems/ezpublish-legacy/blob/master/bin/php/ezsubtreecopy.php#L155)
3. [publish_content.php](https://github.com/ezsystems/ezpublish-legacy/blob/master/bin/php/publish_content.php#L44)

Or with commit performed **before** publishing (as in this PR case)
1. [rssimport.php](https://github.com/ezsystems/ezpublish-legacy/blob/master/cronjobs/rssimport.php#L272)
2. [copysubtree.php](https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/content/copysubtree.php#L236)

From the other hand I found usages with `commit` performed **after** publishing:
1. [eztextfileuser](https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/classes/datatypes/ezuser/eztextfileuser.php#L294)
2. [copy.php](https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/content/copy.php#L96)

andrerom:
Thanks @tomaszmadeyski

gggeek:
@tomaszmadeyski a few notes:

the content/publish operation does have its own transaction (afair of course).

If that is the case, then:

- ezcsvimport: indeed it should use a transaction around the 'create content' section, in case it gets halted just there. I might even have reported this as bug eons ago...
- script ezsubtreecopy: same, might/should get a transaction added for extra safety
- script publish_content.php does no action except publishing => no need for extra transactions

- the case of having a transaction before the publication or encompassing the publication within a transaction would be both valid, and allow for different semantics:
    a) commit after having created an object but no publish => if there is an interruption between the two transactions, or publication fails, you can end up with a consistent db and an object in draft status (this could be verified with a simple test if you want to be 1000% sure: just comment out the publication line and run ezdbintegrity after the fact)
    b) commit after publish => either object gets created and published, or not created at all

While b) seems what you want in most cases, there is the downside that long db transactions harm db concurrency. The long time taken for the publication transaction (there is so much stuff going on in there...) has historically been a big scalability bottleneck for eZ. So a) might be an acceptable tradeoff.

I did not look at the changes you propose in this ticket, nor at the original bug report. But I agree with @andrerom 's sentiment: I have seen more than a couple commits that moved transactions begin/end points getting rolled back because they broke more than they fixed...

gggeek:
Side note: I read the original ticket. It seems that this is a bug happening in the interaction between ez4 and ez5. And that is, pardon my french, known to be buggy already in a few corner cases.

The problem is not the fact that, in ez4, you send a db commit at point A or point B, but rather the fact that ez5 does pile up calls to memcache::clear for execution at the end of the whole process.
( another well known problem is of course that ez4 and ez5 kernels do not share the same db connection. So what one commits, the other might still not see for a while ).

You can find different effects of the same underlying problems f.e. when using the kaliop migration bundle: as soon as you set up ez5 slots, you can not use any more a transaction around multiple content manipulation calls in your ez5 code, as they would be generating fatal errors. I add to add a command line option '--no-transaction' especially to cater to this case.

In other words: I am not a fan of this 'fix'.
To me, a much better fix would be:
1) share db connection between ez4 and ez5 kernel
2) decouple the clearing of the memcache/stash cache from the end of the database transaction

Both of those range 11 on a 1-10 difficulty scale, and risk introducing more bugs than they fix :-(

andrerom:
> as soon as you set up ez5 slots, you can not use any more a transaction ...

this is a bit on the side, but has this been reported, would like to understand that one better.

gggeek:
"... would like to understand that one better ..." => the closest that I can find to an open bug / detailed explanation are the last 2 comments for https://jira.ez.no/browse/EZP-26407, and the original ticket https://github.com/kaliop-uk/ezmigrationbundle/issues/76 . Which is far from being either detailed, clear or concise.
Also,those are only related to the problem I am talking about.
Off the top of my head, one weird case which I had to deal with was: send an object to ezfind for indexation from within an ez5 slot.
There probably were  a couple more, but I forgot the details.
I would have to spend some time, which I don't have at this moment, to back up the original assertion with reproducible examples, and the truth is that it is probably a bit of an exaggeration anyway.

gggeek:
ps: @andrerom thought experiment:
1. $repo->begin
2. alter contentType defintion
3. create content of said class
4. $repo->commit
even without legacy code or slots, if memcache clearing only happens at step 4, step 3 will fail.

pps: stopping the thread hijack here

tomaszmadeyski:
> yes, which is why I'm tempted to say we should rather do an extra cache clear at the end here

It's doable quite easy. In that case `commit` will go where it was originally before 92a89fc4b40332f92400c8ad7d894e16de3e5681 and right after it I will add another call to method to clear the case. It will look something like:

```
    eZContentObject::fixReverseRelations( $objectID, 'restore' );
    
    $db->commit();

    eZContentCacheManager::clearContentCacheIfNeeded( $objectID,  $version->attribute( 'version' ) );
    $module->redirectToView( 'view', array( 'full', $mainNodeID ) );
    return;
```

It may be worse solution performance wise but definitely more secure one. What do you think guys?

andrerom:
> It may be worse solution performance wise but definitely more secure one. What do you think guys?

same, but could you clarify what the remaining case we are trying to solve here now? maybe we can use a more specific event for this to avoid having to clear cache twice in legacy at least, and provide more concrete data for platform to act on.

tomaszmadeyski:
Sure thing: 

Let's imagine class `Article` which has `single_relation` field. It is of `Object relation` type. 

Let's imagine `ItemA` and `ItemB`, both of class `Article`.

`ItemA` has `single_relation` set to `ItemB`. The action performed here is moving `ItemB` to trash and restoring it back from trash. 

Now, we check the value of `single_relation` field of `ItemA`

92a89fc moved `$db->commit` before `eZContentObject::fixReverseRelations` call so it fixed the issue **only in case `ItemB` had `single_relation` field set to `ItemA`**

(it fixed it because `fixReverseRelation` was working on a commited db state already)

So in fact #1306 fixes only one, not so often case.

**The remaining case is the case where `ItemB` has no relation to `ItemA`**. In this case cache for `ItemB` is cleared as a result of publishing `ItemA` but since restoration is not commited yet it won't be warmed up properly: when you check value of `single_relation` of `ItemA` it will show `null`

tomaszmadeyski:
After a chat with @andrerom I reverted the position of `$db->commit()` to state before #1306 and added additional call to clear the cache after commit is performed.
I also added a flag `clearCacheIfEnabled` to `eZContentObject::fixReverseRelations` method to avoid double cache clearing while restoring

kernel/content/restore.php
```diff
@@ -134,7 +134,6 @@
     $version->store();
 
     $object->restoreObjectAttributes();
-
```
andrerom: I guess we can skip this, and if we want to reduce the changes of the two PR's in combination there should afaik be a extra new line before `$db->commit();` :P

PR #1311, opened 2017-08-01
EZP-27737: Load correct groups/roles in ezuser

wizhippo:
> https://jira.ez.no/browse/EZP-27737

The fixes a regression after https://github.com/ezsystems/ezpublish-legacy/commit/134b652e66a399743ba158bf06c101c8abca7bd4

### DIFF
--- a/kernel/classes/datatypes/ezuser/ezuser.php
+++ b/kernel/classes/datatypes/ezuser/ezuser.php
@@ -1420,11 +1420,12 @@ WHERE user_id = '" . $userID . "' AND
         // At the end of this method flag is set to previous value again
         $previousCachingState = $user->setCachingEnabled( false );
         // user groups list (session: eZUserGroupsCache)
-        $data['groups'] = $user->groups();
+        $groups = $user->groups(); 
+        $data['groups'] = $groups;
 
         // role list (session: eZRoleIDList)
-        $data['groups'][] = $userId;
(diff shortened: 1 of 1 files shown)


kernel/classes/datatypes/ezuser/ezuser.php
```diff
@@ -1423,8 +1423,8 @@ private static function generateUserCacheData( $userId )
         $data['groups'] = $user->groups();
 
         // role list (session: eZRoleIDList)
-        $groups[] = $userId;
-        $data['roles'] = eZRole::fetchIDListByUser( $groups );
+        $data['groups'][] = $userId;
```
adamwojs: This method returns `$data` as result. In original behaviour  `$data['groups']` wasn't contained `$userId` at the end. So I think that this code should looks like as follows:

```php
$groups = $user->groups(); 
$data['groups'] = $groups;
$groups[] = $userId;

$data['roles'] = eZRole::fetchIDListByUser($groups);
```

wizhippo:
Not sure why eZDFSClusterStaleCacheTest::testStaleCache is failing.  Don't think is has anything to do with this commit.

adamwojs:
> Don't think is has anything to do with this commit.

@wizhippo Other PR are also failing this test, so don't worry about this 😉

andrerom:
Merged, thanks @wizhippo!

PR #1312, opened 2017-08-02
EZP-27735 Solr index not updated immediately after change content priority

kmadejski:
JIRA issue: [https://jira.ez.no/browse/EZP-27735](https://jira.ez.no/browse/EZP-27735)

# Description
First change of content priority was incorrectly indexed in Solr. Instead of new value, default 0 value was indexed.

### DIFF
--- a/kernel/content/ezcontentoperationcollection.php
+++ b/kernel/content/ezcontentoperationcollection.php
@@ -1244,6 +1244,7 @@ class eZContentOperationCollection
              $db->commit();
              if ( !eZSearch::getEngine() instanceof eZSearchEngine )
              {
+                 eZContentCacheManager::clearContentCacheIfNeeded( $objectIDs );
                  foreach ( $objectIDs as $objectID )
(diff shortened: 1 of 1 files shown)


pkamps:
The code changes are not affecting how solr gets update. The title is misleading.

The code changes clear the content cache for all objects with new priority (function updatePriority). It's probably a good idea to clear the content cache for those nodes. For example if the node priority is listed on the full view of a node.

kmadejski:
@pkamps thank you for your feedback. You are partially right, but PR title was created based on JIRA issue. The issue was discovered in relation with Solr and concerned this unwanted behavior which affected only Solr search.

andrerom:
Title reflect the issue in hybrid setup, where this solves an issue where solr index is not updated since persistence cache is not cleared before the indexing, this was the context where it was reported. Your suggestion for title @pkamps is right, but mainly makes sense in a pure legacy context.

PR #1313, opened 2017-08-02
[Composer] Specify conflict with kernel older then 6.11

andrerom:
As part of [2017.08](https://github.com/ezsystems/ezpublish-legacy/milestone/1) milestone we are currently merging a few new features to keep legacy compatible with latests version of eZ Platform, which will cause issues if they it is used together with older kernel installs like 6.7 and 5.4.

This was suggested by @emodric in [ongoing](https://github.com/ezsystems/LegacyBridge/pull/103) work to simplify legacy bridge install, where simplifying also tries to avoid people ending up in non working package combinations.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -61,5 +61,8 @@
     "require-dev": {
         "phpunit/phpunit": "3.7.*",
         "zetacomponents/php-generator": "~1.1"
+    },
+    "conflict": {
+        "ezsystems/ezpublish-kernel": "<6.11@dev || >=2014.11"
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -64,5 +64,10 @@
     },
     "conflict": {
         "ezsystems/ezpublish-kernel": "<6.11@dev || >=2014.11"
+    },
+    "extra": {
+        "branch-alias": {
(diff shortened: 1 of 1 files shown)


emodric:
`>=2014.11` should also be added to conflicts, probably.

andrerom:
True, but unsure if that should be a `||` or ` ` _(and)_ or `,`, the doc on this one is rather hard to read pre caffeinated: https://getcomposer.org/doc/04-schema.md#conflict

emodric:
Looks like `||` to me. As in `<6.11@dev || >=2014.11`.

stef03:
How can I use ezpublish-legacy with ezpublish-kernel 6.5? Till today I was able to use "dev-master".

andrerom:
@stef03 any reason why you use dev-master _(unstable dev version)_ as opposed to either 2014.11, or community forks like

There will be a 2017.08 release coming very soon that will replace both, but as stated here it will require kernel 6.11 because of changes to among other things relation field/data type.

stef03:
@andrerom Thank you for helping us. We are now using the netgen community fork with Version "v2017.07.0".

PR #1315, opened 2017-08-15
Fix EZP-27814: Support MySQL error codes in DB exceptions

peterkeung:
Bug report: https://jira.ez.no/browse/EZP-27814

Since eZ Publish 4.5, we can catch DB exceptions: https://github.com/ezsystems/ezpublish-legacy/blob/master/doc/features/4.5/ezdb_exception_based_error_handling.txt

However, the error code in the DB exception is always 0. This pull request logs the MySQL error code in the exception code so that you can specifically catch error codes (such as deadlocks) and act accordingly (such as retrying the transaction).

One of the reasons this is important is that the legacy API is prone to DB deadlock errors on concurrent object creation.

You can reproduce this quite easily by running multiple versions of this script at the same time:

```
<?php
for( $i = 0; $i < 100; ++$i )
{
    $params = array();
    $params['class_identifier'] = 'folder';
    $params['parent_node_id'] = 62;
    $params['attributes'] = array( 'name' => 'Test folder script 1' );
 
    $contentObject = eZContentFunctions::createAndPublishObject( $params );
}
```

By adding the DB exception code, we can catch a deadlock error and retry the creation rather than having the script crash.

### DIFF
--- a/kernel/private/classes/exceptions/database/exception.php
+++ b/kernel/private/classes/exceptions/database/exception.php
@@ -17,5 +17,30 @@
  */
 class eZDBException extends ezcBaseException
 {
+    /**
+     * Original message, before escaping
+     */
+    public $originalMessage;
+
+    /**
+     * Constructs a new eZDBException with $message and $code
+     *
+     * @param string $message
+     * @param int $code
+     */
+    public function __construct( $message, $code = 0 )
+    {
+        $this->originalMessage = $message;
+        $this->code = $code;
+
+        if ( php_sapi_name() == 'cli' )
+        {
+            $this->message = $message;
+        }
+        else
+        {
+            $this->message = htmlspecialchars( $message, ENT_QUOTES, 'UTF-8' );
+        }
+    }
 }
 ?>
--- a/lib/ezdb/classes/ezmysqlidb.php
+++ b/lib/ezdb/classes/ezmysqlidb.php
@@ -402,7 +402,8 @@ class eZMySQLiDB extends eZDBInterface
             }
             else
             {
-                $errorMessage = 'Query error (' . mysqli_errno( $connection ) . '): ' . mysqli_error( $connection ) . '. Query: ' . $sql;
+                $this->setError();
+                $errorMessage = 'Query error (' . $this->ErrorNumber . '): ' . $this->ErrorMessage . '. Query: ' . $sql;
                 eZDebug::writeError( $errorMessage, __CLASS__  );
                 $oldRecordError = $this->RecordError;
                 // Turn off error handling while we unlock


kernel/private/classes/exceptions/database/exception.php
```diff
@@ -15,7 +15,31 @@
  * @version //autogentag//
  * @package kernel
  */
-class eZDBException extends ezcBaseException
+class eZDBException extends Exception
```
andrerom: break

peterkeung: What's here now is essentially a copy of ezcBaseException, but with support for the error code / number. Thus, the intention is that there is no BC break at all.

emodric: It's a break if someone typehinted against `ezcBaseExtension`.

Can you not implement a custom base exception extending from `ezcBaseExtension` with added features?

pkamps: It's unlikely somebody would typehint against 'ezcBaseException'. It's much more likely that somebody would typehint against 'eZDBException'

emodric: Sure, but that doesn't change the fact that it's still a BC break which IMO should not be made lightly.

peterkeung: OK, I've rewritten it so that we continue to extend ezcBaseException but just set the $message and $code properties, since the ezcBaseException constructor removed $code support.

kernel/private/classes/exceptions/database/exception.php
```diff
@@ -17,5 +17,30 @@
...
+     */
+    public $originalMessage;
+
+    /**
+     * Constructs a new eZDBException with $message and $code
+     *
+     * @param string $message
+     * @param int $code
+     */
+    public function __construct( $message, $code = 0 )
```
emodric: `$message` should also have a default value (an empty string) to preserve BC.

kernel/private/classes/exceptions/database/exception.php
```diff
@@ -17,5 +17,30 @@
...
+        $this->originalMessage = $message;
+        $this->code = $code;
+
+        if ( php_sapi_name() == 'cli' )
+        {
+            $this->message = $message;
+        }
+        else
+        {
+            $this->message = htmlspecialchars( $message );
```
emodric: This should probably be `htmlspecialchars($message, ENT_QUOTES, 'UTF-8')` to properly encode quotes and since UTF-8 is the default charset only since PHP 5.4.

emodric:
Apart from the comment about `htmlspecialchars`, this looks good :+1:

kernel/private/classes/exceptions/database/exception.php
peterkeung: It was originally just a copy of `ezcBaseException`, but I've made the change now :)

peterkeung:
About the constants, in the context of MySQL (https://dev.mysql.com/doc/refman/5.7/en/error-messages-server.html), I don't think there is such a list of them in PHP.

moismailzai:
+1

dfearnley:
+1

ernestob:
This looks good 👍

andrerom:
@peterkeung Note: Didn't see this before after merge, but it might be a mistake not to call  `parent::__construct( $message )` as [Exception::__construct()](https://github.com/php/php-src/blob/cd953269d3d486f775f1935731b1d6d44f12a350/ext/spl/spl.php#L284-L293) then won't get called.

peterkeung:
@andrerom That was the entire challenge of preserving the use of ezcBaseException, but I think we are good here, because:
* The exception is thrown
* You can catch eZDBException, ezcBaseException, or Exception
* By setting the $this->message and $this->code properties, we are making the necessary information available to be acted upon

Thus, we don't need to call parent::__construct( $message) unless I'm missing something?

andrerom:
I was thinking about these parts:
```php
		$this->file = __FILE__; // of throw clause
		$this->line = __LINE__; // of throw clause
		$this->trace = debug_backtrace();
		$this->string = StringFormat($this);
```

peterkeung:
@andrerom (Disclaimer: this is beyond my exception knowledge, so I apologize if I have a fundamental misunderstanding.) In my tests, I still got the expected results with $e->getFile(), $e->getLine(), $e->getTrace(), $e->getTraceAsString() in the caught exception object. We couldn't just move those lines into eZDBException or we'd get wrong info.

andrerom:
>  In my tests, I still got the expected results with $e->getFile(), $e->getLine(), $e->getTrace(), $e->getTraceAsString() in the caught exception object. 

ok then we are good, my doubt was if these would be wrong now.

> We couldn't just move those lines into eZDBException or we'd get wrong info.

was not implying that, only that we'd somehow call  Exception::__construct().
But if the info is right, then no need. done.

kernel/private/classes/exceptions/database/exception.php
```diff
@@ -17,5 +17,30 @@
...
+        $this->originalMessage = $message;
+        $this->code = $code;
+
+        if ( php_sapi_name() == 'cli' )
+        {
+            $this->message = $message;
+        }
+        else
+        {
+            $this->message = htmlspecialchars( $message, ENT_QUOTES, 'UTF-8' );
```
gggeek: Nitpick: it is bad practice to html-encode data on data-creation. Data should be escaped when it is displayed.

PR #1317, opened 2017-08-17
Font rewrite rule

dfearnley:
It is pretty common that our design extensions have web fonts these days.
The default rewrite rules that come in .htaccess_root or the vhost demo file on each eZ version
do not have an exception for a fonts folder yet, which results in the web fonts breaking until it is manually added after deployment.


dfearnley:
Pull request was not properly created.

PR #1318, opened 2017-08-28
Fix EZP-27803: Wrong URLs in Treemenu if Platform is installed in a sub-directory

blankse:
JIRA: https://jira.ez.no/browse/EZP-27803


emodric:
Looks good :) I suppose this has to do with front controller being named `app.php` ?

blankse:
Yes, eZSys must have the right front controller `app.php` to generate right URLs.
The solution is the same as in ezpkernelweb.php:
https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/private/classes/ezpkernelweb.php#L229

PR #1319, opened 2017-09-02
Fix EZP-28165: Incorrect increment of object_count in ezsearch_word

gabrielfin:
When updating object_count in ezsearch_word, eZSearchEngine splits the list of words in chunks of 500. But in every cycle of the loop not only the current chunk but all of the previous chunks are incremented. Which means that if there are, lets say 3 chunks, the first chunk gets an increment of +3, the second of +2, and the third of +1.

Steps to reproduce:
1. Create a new object with an attribute that has (plenty) more than 500 indexable words.
2. Pick any word that is mentioned at the beginning of that attribute.
3. Publish the object
4. Find that word in ezsearch_word and watch the object_count value.
5. Edit the object created in step 1 and publish it without modifying it.
6. Find again the word in ezsearch_word and watch the object_count value.

Result:
The object_count has a larger value than it should.

Fix:
Simply move `$wordIDArray` from outside the `for` to inside, so that it is reset on every loop. I also renamed the variable to `$wordIDArrayChuck`, to make it consistent with `$wordArrayChuck`, but that's not necessary.

### DIFF
--- a/kernel/search/plugins/ezsearchengine/ezsearchengine.php
+++ b/kernel/search/plugins/ezsearchengine/ezsearchengine.php
@@ -168,7 +168,6 @@ class eZSearchEngine implements ezpSearchEngine
         $trans = eZCharTransform::instance();
 
         $wordCount = count( $indexArrayOnlyWords );
-        $wordIDArray = array();
         $wordArray = array();
         // store the words in the index and remember the ID
         $dbName = $db->databaseName();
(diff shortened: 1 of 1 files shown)


kernel/search/plugins/ezsearchengine/ezsearchengine.php
```diff
@@ -185,16 +184,17 @@ function buildWordIDArray( $indexArrayOnlyWords )
                 // Build a has of the existing words
                 $wordResCount = count( $wordRes );
                 $existingWordArray = array();
+                $wordIDArrayChuck = array();
```
andrerom: could we just move init of `$wordIDArray` here to avoid needing to change lines below?

gabrielfin:
Done.

andrerom:
Thanks! Could you btw create [JIRA](jira.ez.no) issue for this? Affected version will be 2017.10

gabrielfin:
Done: https://jira.ez.no/browse/EZP-28165

PR #1321, opened 2017-09-25
[Travis] Setup PHP7 testing on travis, bump requriment to PHP 5.5

andrerom:
Notes:
- PHP 5.3 no longer supported on latests version of travis env, so bumping version to PHP 5.5 which is the lowest version I know currently maintained by up-to-date linux distros.
  - NOTE: Minimum version for use with eZ Platform is PHP 5.6, but PHP 7.x is recommended
- This will most likly fail becasue of PHPUnit version, so might need some help on that part to get this to pass


tests/tests/kernel/private/rest/request_parser_data/get_with_https.data.exp
```diff
@@ -31,32 +31,6 @@ ezpRestRequest::__set_state(array(
   ),
    'body' => '',
    'files' => NULL,
-   'accept' =>
```
emodric: Why is this removed?

As far as I can see, builds mostly fail because this is missing.

andrerom: I know, I'll need to handle this differentially. issue was PHP >= 7.1 has the headers (types, charsets, encodings) in different order for some reason. And I thought I would get away with having to normalize it this way _(as it is afaik not what the test cases is caring about anyway)_.

But, I'll probably normalize it once I have time.

emodric: If the test case doesn't care about `accept`, why not just `unset` it before asserting? :)

andrerom:
ping @emodric _The header ordering is a zeta issue with php7, reported in link on the specific commit. As said above not the main part of the test here so the fix here just makes sure ordering is always correct._

emodric:
Not sure about the change in `attribute_edit.php` as I'm less familiar with the code there, but I trust you :P Looks good otherwise :+1:

andrerom:
> Not sure about the change in attribute_edit.php

It's actually a bug fix kind of, so I'll create JIRA issue for it. The code was causing error and seemed to be travis being configured with this set to 0, which the code there was not made to handle.

kernel/content/attribute_edit.php
```diff
@@ -219,35 +219,40 @@
...
-        case 'k':
-            $postMaxSizeBytes *= 1024;
-            $postMaxSizeUnit = 'Kb';
-            break;
-    }
-    if ( (int)$_SERVER['CONTENT_LENGTH'] > $postMaxSizeBytes &&  // This is not 100% acurrate as $_SERVER['CONTENT_LENGTH'] doesn't only count post data but also other things
-        count( $_POST ) === 0 )                                 // Therefore we also check if request got no post variables.
+
+    // As of PHP 5.3.2 it is possible to configure unlimited post_max_size, further checking
+    if ($postMaxSize == 0)
```
blankse: Should be `$postMaxSize != 0` or?

andrerom: indeed, hmm, why did that pass 😕

PR #1322, opened 2017-09-26
EZP-24744 - Increase password security

glye:
> https://jira.ez.no/browse/EZP-24744
> Based on https://github.com/netgen/ezpublish-legacy/commit/7d5bd88cb7ceba3e3ec1c349786e1022d375d38f by @emodric 
> which is based on https://github.com/ezsystems/ezpublish-legacy/pull/1195 by @pedroresende 
> Same for new stack: https://github.com/ezsystems/ezpublish-kernel/pull/2095
> Ready for review

Added PASSWORD_HASH_PHP_DEFAULT in addition to PASSWORD_HASH_BCRYPT
Set PASSWORD_HASH_PHP_DEFAULT as default


glye:
Test failure is not my fault - it cannot download php-5.3.tar.bz2

update/database/postgresql/7.0/dbupdate-5.4.0-to-7.0.0.sql
```diff
@@ -0,0 +1,4 @@
+UPDATE ezsite_data SET value='7.0.0' WHERE name='ezpublish-version';
```
emodric: Should this be 6.12?

share/db_data.dba
```diff
@@ -23048,16 +23048,16 @@ keywords=cms, publish, e-commerce, content management, development framework',
         0 => '10',
         1 => 'nospam@ez.no',
         2 => 'anonymous',
-        3 => '4e6f6184135228ccd45f8233d72a0363',
-        4 => '2',
+        3 => '$2y$10$wUrvpYTs9KZ0mtB9Gzr.G.nk6wGhPYzcH3jl9Tk5DtxJN8povoQi2',
+        4 => '6',
```
emodric: I'd like to see this consistent with the default value in config (`php_default`)

update/database/postgresql/7.0/dbupdate-5.4.0-to-7.0.0.sql
andrerom: yes

glye: @emodric @andrerom You mean set to `6.12.0` in file `*/6.12/dbupdate-5.4.0-to-6.12.0.sql`, and same for mysql, right?

emodric: Yep, exactly.

share/db_data.dba
glye: @emodric @andrerom Ok. Do you know which password should be used for the anonymous hash? If it matters at all... :)

emodric: IIRC, I generated those hashes for `publish` password, for all accounts.

glye: Nope, I checked and that's not it. But I can as well use it, I don't think it matters what it is.

emodric: You're right. It's `anonymous` for anonymous user and `publish` for admin user.

glye: That doesn't fit the existing data either. I used `publish` for both.

glye:
@emodric Changed dbupdate version number, and updated hashes in db data to php_default type.

emodric:
A test is failing now, it seems, so once it's fixed, :+1:

glye:
Hm.. this test is kind of obsolete now, and if we're going to keep it we'll have to keep a db entry with md5_user hash, too.

emodric:
Can we update it so it can use `php_default` instead of `md5_user` ?

glye:
The test makes no sense with php_default, as it doesn't use usernames at all in the hash.
Looking at the code, it should actually work, though. I'll dig some more. 👷

glye:
Reverting db data update, to verify that the problem is indeed there. (Not reproducible locally)

composer.json
```diff
@@ -57,14 +57,15 @@
         "zetacomponents/persistent-object": "~1.8",
         "zetacomponents/signal-slot": "~1.2",
         "zetacomponents/system-information": "~1.1",
-        "zetacomponents/webdav": "~1.1"
+        "zetacomponents/webdav": "~1.1",
+        "ircmaxell/password-compat": "^1.0"
```
andrerom: hmm, as we now require php 5.5 as of cf005524485c89f82e6d24e979dfc26d659e45cd, we probably don't need this anymore.

settings/site.ini
```diff
@@ -562,10 +562,12 @@ UserCreatorID=14
 # md5_user generates password hash from user and password.
 # md5_site generates password hash from site, user and password
 # plaintext does not generate a hash but has the password as it is, this is not
-# recommended since it is a huge security risc.
+# recommended since it is a huge security risk.
+# bcrypt generates hash from password only
+# php_default uses whichever method PHP deems appropriate (currently bcrypt, as of PHP 7)
```
andrerom: `as of PHP 7` => `as of PHP 5.5`: http://php.net/password_hash

tests/tests/kernel/datatypes/ezuser/ezuser_test.php
```diff
@@ -59,9 +59,14 @@ public function testPasswordHashSamePasswordToUser()
...
             $this->fail( "User {$this->username} is not in database.");
         }
-        // Not used in this test
         $passwordHashMD5Password = $rows[0]['password_hash'];
 
+        // The value that is expected to be saved in the ezuser table with the HashType to md5_password
+        $hashMD5Expected = md5( $this->password );
+
+        // Verify that the 2 password hashes saved above are the same
+        $this->assertEquals( $hashMD5Expected, $passwordHashMD5Password );
```
andrerom: I would kind of expect this to now fail as password in db is blowfish and not md5, rest afterwards I would expect to pass. Strange, no local reproduction of this with mysql?

glye: All logins here run with `UpdateHash=true`, so at this point the password in db has been changed to `md5_password`. What I don't understand is why it isn't changed to `md5_user` during the next login, which is causing the failing tests. I also have no idea why https://github.com/ezsystems/ezpublish-legacy/pull/1322/commits/7b0136add5f9a06f6e6f819646238a0d3fc15a59 is what triggers it.

Can't reproduce it locally, no.

composer.json
glye: Right! Removed.

settings/site.ini
glye: Fixed, thanks.

glye:
Passing, finally! I'll clean it up and then we're done.

emodric:
@glye So what the root cause of the failure?

glye:
Cleaned up commits, this is how I would like to merge it, when it passes.

@emodric Test data having the same password and the same email address for admin and anon users.

emodric:
> @emodric Test data having the same password and the same email address for admin and anon users.

Huh, so where is it fixed? Can't see anything in the diff.

glye:
@emodric I changed the password for anonymous. Hard to see the change in the hash, since it's changing from MD5 anyway :)

emodric:
> @emodric I changed the password for anonymous. Hard to see the change in the hash, since it's changing from MD5 anyway :)

Hah, okay :) So which one is it now? `anonymous` was before, right?

glye:
It is `anonymous` now, was something else before. Never found out what.

emodric:
:+1:

emodric:
@glye If you wish, you can change the initial commit to give @pedroresende credit instead of me, it's his code to begin with.

glye:
@emodric It's green, not sure I dare to touch it anymore 😁

emodric:
Okay, it's up to you :D

glye:
Re-author-ified now :)

emodric:
@andrerom I think my version was because of `PASSWORD_BCRYPT` => `PASSWORD_HASH_BCRYPT`

andrerom:
@emodric true

emodric:
Yep! Merge away!

emodric:
Hm... Changing the password for a user with old hash types does not update the hash type to the new value.

@andrerom  @glye Do you guys know if this is by design in legacy?

PR #1323, opened 2017-09-28
EZP-27975: Support post_max_size value set to 0 during content/publish in UI

andrerom:
Correct logic of c95ee808ea836a2f3d3b83876bafbfb208dde4a0

### DIFF
--- a/kernel/content/attribute_edit.php
+++ b/kernel/content/attribute_edit.php
@@ -219,11 +219,11 @@ if ( $_SERVER['REQUEST_METHOD'] === 'POST' )
 {
     // Need to detect if post_max_size has been reached. If so, all post variables are gone...
     $postMaxSize = trim( ini_get( 'post_max_size' ) );
+    $postMaxSizeBytes = (int)$postMaxSize;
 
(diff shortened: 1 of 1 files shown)


kernel/content/attribute_edit.php
```diff
@@ -221,7 +221,7 @@
     $postMaxSize = trim( ini_get( 'post_max_size' ) );
 
     // As of PHP 5.3.2 it is possible to configure unlimited post_max_size, further checking
-    if ($postMaxSize == 0)
+    if ($postMaxSize != 0)
     {
         $postMaxSizeBytes = $postMaxSize;
```
blankse: This could help for the failing test:
`$postMaxSizeBytes = (int)$postMaxSize;`

andrerom: 👍  , but I'll first var dump what the hell is configured here. maybe it's 0b or something, but have to know 😆

blankse: It fails in line 239. So it is something with 'm' at the end :)

andrerom:
thanks @blankse

PR #1324, opened 2017-09-28
EZP-27993: Support quality override by image alias

peterkeung:
https://jira.ez.no/browse/EZP-27993

In eZ Publish legacy, we can set a global image quality in image.ini, but cannot override this per alias.
This is the reverse problem as eZ Platform / new stack, where you can set the quality by alias but cannot set a global value (https://github.com/liip/LiipImagineBundle/pull/909).

This way, you can set something like this in image.ini to override the global setting:

```
[medium]
Quality[]
Quality[]=image/jpeg;80
```

### DIFF
--- a/lib/ezimage/classes/ezimagemanager.php
+++ b/lib/ezimage/classes/ezimagemanager.php
@@ -98,6 +98,7 @@ class eZImageManager
         $this->MIMETypeSettingsMap = array();
         $this->QualityValues = array();
         $this->QualityValueMap = array();
+        $this->QualityValueMapOverride = array();
 
         $ini = eZINI::instance( 'image.ini' );
         $this->TemporaryImageDirPath = eZSys::cacheDirectory() . '/' . $ini->variable( 'FileSettings', 'TemporaryDir' );
@@ -177,6 +178,7 @@ class eZImageManager
             $alias = array( 'name' => 'original',
                             'reference' => false,
                             'mime_type' => false,
+                            'quality' => false,
                             'filters' => array() );
             $alias['alias_key'] = $this->createImageAliasKey( $alias );
             $aliasList['original'] = $alias;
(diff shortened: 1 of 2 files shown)


settings/image.ini
```diff
@@ -76,6 +76,11 @@ AliasList[]=rss
 # to use image/jpeg for screenshot but rather image/png or image/gif.
 # MIMEType=image/png
 #
+# Override quality by MIME type for this alias only
+# This follows the pattern from the global [MIMETypeSettings] settings 
+# Quality[]
+# Quality[]=image/jpeg;80
```
emodric: In PR description you mention `Quality[image/jpeg]=80` as format. This is the correct one as far as I can see?

lib/ezimage/classes/ezimagemanager.php
```diff
@@ -1059,6 +1084,17 @@ function convert( $sourceMimeData, &$destinationMimeData, $aliasName = false, $p
                 {
                     eZMimeType::changeMIMEType( $destinationMimeData, $alias['mime_type'] );
                 }
+                $qualityOverride = $alias['quality'];
+                if ( $qualityOverride )
+                {
+                    foreach ( $qualityOverride as $qualityOverrideValue )
+                    {
+                        $elements = explode( ';', $qualityOverrideValue );
+                        $mimeType = $elements[0];
```
emodric: Maybe do a basic check if `$elements[0]` and `$elements[1]` exist before using them?

emodric:
Useful feature :+1: 

I'm wondering about why use the override values which are constantly reset and filled for every image conversion process? I guess there's no other way if we want to keep `eZImageManager` interface intact?

settings/image.ini
peterkeung: Sorry, I've fixed the PR description now! Good catch. I got distracted by the format of the internal map.

lib/ezimage/classes/ezimagemanager.php
peterkeung: I was copying readMIMETypeQualitySettingFromINI() on line 559, so it's consistent to expect the user to follow the right format. Could harden this in a follow-up PR though.

peterkeung:
Right, this was what I came up with without having to send the alias information to each individual conversion handler. I can certainly change it if someone thinks of a way to do this per alias while maintaining the split in concerns between eZImageManager and the conversion handlers.

peterkeung:
Ping @alexgorski

alexgorski:
Very useful feature requested by amnh-digital.

thiagocamposviana:
Looks good to me
Basically before we could only set the quality by mimetype, now it is possible to define quality by alias, which is very important, also defining quality only by mimtype is so limited it is not very helpful.

ernestob:
Besides the comment from @emodric about testing the values of the quality array, this looks good.

This is a must have feature. 👍

benkmugo:
Great addition for sites that have a lot of aliases where quality needs to be controlled better due to alias size differences e.g. huge banners with high-quality needs vs. thumbnails that can be lower quality.

We've got at least a handful of clients where this update would mean a big improvement.

+1 from me

andrerom:
@crevillo you are also good with this?

crevillo:
yes! I am, sorry, i didn't see this notification. @peterkeung just pinged me about.

andrerom:
Thanks @crevillo and @emodric for the reviews, and thansk for the work here @peterkeung!

PR #1326, opened 2017-10-03
EZP-28010: Update password hash type to current default when changing the user password

emodric:
> https://jira.ez.no/browse/EZP-28010

Now that #1322 is implemented, it would be wise to update the hash to the new default whenever a user changes the password.

### DIFF
--- a/kernel/user/ezuseroperationcollection.php
+++ b/kernel/user/ezuseroperationcollection.php
@@ -311,10 +311,11 @@ class eZUserOperationCollection
         if ( $user instanceof eZUser )
         {
             $login   = $user->attribute( 'login' );
-            $type    = $user->attribute( 'password_hash_type' );
+            $type    = eZUser::hashType();
             $site    = $user->site();
             $newHash = $user->createHash( $login, $newPassword, $site, $type );
             $user->setAttribute( 'password_hash', $newHash );
+            $user->setAttribute( 'password_hash_type', $type );
             $user->store();
 
             // "Draft" must be in sync with the PersistentObject


andrerom:
+1

For Platform it seems to be opposite, we are missing code to update hash on login afaik _(only checked user service and not the login logic in MVC space, so I might be wrong here)_

emodric:
@andrerom Issue created.

I don't remember that it does. But, legacy stack also doesn't update the hash on user login AFAIK.

andrerom:
In legacy there is a setting for it as far as I can see: https://github.com/ezsystems/ezpublish-legacy/blob/18109cbdf500f48c9f736d2d2139aeeafa6e181d/kernel/classes/datatypes/ezuser/ezuser.php#L868

Enabled by default: https://github.com/ezsystems/ezpublish-legacy/blob/430df048919f840d6c5c67a068073a0cc616e7b9/settings/site.ini#L578

For me this makes sense, the system knows the password during login, so can do gradual migration of passwords to imrpvoe password strenght. 

@glye This opens an missing feature for future proofing btw, once PHP updates default hash type the system will no longer detect this on login and update hash, since all eZ knows about is that default_php type is set.

andrerom:
@emodric I was about to suggest we update this PR to take into account `UpdateHash` setting here as well, however that will need doc updates, and frankly more secure to leave this as you have suggested it here. Update on login is maybe tiny bit more natural to have a setting for, even if I would perosnally skip that in Platform if we add that logic there /cc @glye

emodric:
> I was about to suggest we update this PR to take into account UpdateHash setting here as well, however that will need doc updates, and frankly more secure to leave this as you have suggested it here. 

I can maybe add a separate config with a default value set to `true`? That way security wouldn't suffer immediately and later would depend purely on user choice.

> Update on login is maybe tiny bit more natural to have a setting for, even if I would perosnally skip that in Platform if we add that logic there

I would too. After all it's an implementation detail.

glye:
I have tested UpdateHash and it does indeed update passwords on login. @andrerom good point about future changes to default hash type, will need a code change.

andrerom:
> I can maybe add a separate config with a default value set to true? 

lets not add that before someone actually surfaces the need for it, default hash type is simply taken into account on password update just like on user creation, and from what we now right now that makes sense to always do for system consistency.

andrerom:
@glye Are you aware about any existing doc that will be affected by this? Otherwise I'll just opt for clarifying this in release notes for 2017.10.

glye:
@andrerom Nope. Ok, might as well do that, yes.

emodric:
Thanks @andrerom @glye !

PR #1331, opened 2017-11-15
datetime.ini: Inline doc added. Better examples

pkamps:
I copied the documentation from the ez site into the ini file. Additionally I added a few more useful examples for dates and times.

### DIFF
--- a/settings/datetime.ini
+++ b/settings/datetime.ini
@@ -46,6 +46,7 @@
 
 [ClassSettings]
 Formats[]
+Formats[all]=%a, %A, %d, %D, %F, %g, %G, %h, %H, %i, %j, %l, %m, %M, %n, %O, %s, %T, %U, %w, %W, %Y, %y, %z, %Z, %r, %c
 
 Formats[short_date]=%j.%n.%Y
 Formats[long_date]=%j %F %Y
--- a/settings/datetime.ini
+++ b/settings/datetime.ini
@@ -1,14 +1,56 @@
 #?ini charset="utf-8"?
 # eZ Publish configuration file for date and time handling.
 #
-# NOTE: It is not recommended to edit this files directly, instead
-#       a file in override should be created for setting the
-#       values that is required for your site. Either create
-#       a file called settings/override/datetime.ini.append or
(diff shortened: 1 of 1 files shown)


settings/datetime.ini
```diff
@@ -1,14 +1,56 @@
...
-# NOTE: It is not recommended to edit this files directly, instead
-#       a file in override should be created for setting the
-#       values that is required for your site. Either create
-#       a file called settings/override/datetime.ini.append or
-#       settings/override/datetime.ini.append.php for more security
-#       in non-virtualhost modes (the .php file may already be present
-#       and can be used for this purpose).
+# NOTE:  It is not recommended to edit this files directly, instead
+#        a file in override should be created for setting the
+#        values that is required for your site. Either create
```
crevillo: `is` shouldn't be `are` here? (sorry about my english...)

settings/datetime.ini
```diff
@@ -1,14 +1,56 @@
...
+#  %W         41                               ISO-8601 week number of year, weeks starting on Monday.
+#  %Y         2003                             A full numeric representation of a year, 4 digits.
+#  %y         03                               A two digit representation of a year.
+#  %z         280                              The day of the year.
+#  %Z         -18000                           Timezone offset in seconds. The offset for timezones west of UTC is
+#                                              always negative, and for those east of UTC is always positive.
 
 [ClassSettings]
 Formats[]
-Formats[all]=%a, %A, %d, %D, %F, %g, %G, %h, %H, %i, %j, %l, %m, %M, %n, %O, %s, %T, %U, %w, %W, %Y, %y, %z, %Z, %r, %c
```
crevillo: shouldn't we keep this and probably make it deprecated before finally remove it?

settings/datetime.ini
gggeek: also 'this files' that is wrong :-D

settings/datetime.ini
pkamps: I just cannot imagine that somebody is using it. It's just useless.

emodric: It's still a breaking change. It would be good to stick to the PR description :)

PR #1334, opened 2017-11-29
EZP-28214: Password hash silently defaults to MD5

glye:
> Fix https://jira.ez.no/browse/EZP-28214 for legacy
> For new stack, see https://github.com/ezsystems/ezpublish-kernel/pull/2171
> Sent to QA

Invalid password hash type should not be ignored: Log an error if an invalid password hash type is specified. Fall back to the default hash type (currently bcrypt).

- [x] Add new constant for the default hash type, so that on future change this can be updated in a single place.

### DIFF
--- a/kernel/classes/datatypes/ezuser/ezuser.php
+++ b/kernel/classes/datatypes/ezuser/ezuser.php
@@ -32,6 +32,10 @@ class eZUser extends eZPersistentObject
     /// Passwords in PHP default format
     const PASSWORD_HASH_PHP_DEFAULT = 7;
 
+    /// Default password hashing algorithm, used in case of invalid configuration or usage.
+    /// Update this if support for better algorithms is added.
+    const DEFAULT_PASSWORD_HASH = self::PASSWORD_HASH_PHP_DEFAULT;
+
     /**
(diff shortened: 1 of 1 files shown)


kernel/classes/datatypes/ezuser/ezuser.php
```diff
@@ -32,6 +32,10 @@ class eZUser extends eZPersistentObject
     /// Passwords in PHP default format
     const PASSWORD_HASH_PHP_DEFAULT = 7;
 
+    /// Default password hashing algorithm, used in case of invalid configuration or usage.
+    /// Update this if support for better algorithms is added.
+    const DEFAULT_PASSWORD_HASH = self::PASSWORD_HASH_MD5_USER;
```
emodric: Why not `PASSWORD_HASH_BCRYPT` ?

glye: @emodric Artifact from earlier iteration. Should be `PASSWORD_HASH_PHP_DEFAULT`. I'll correct it. Thanks!

emodric:
@andrerom Should this go in?

PR #1337, opened 2017-12-19
Bump master to 2017.12

emodric:


andrerom:
Also need to bump lib/versions.php

emodric:
@andrerom Done.

PR #1343, opened 2018-01-12
EZP-28404: Execution of "Approve" workflow event type doesn't clear SPI cache

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-28404

## Description 

This PR adds a missing dispatching of `content/cache/version` event after execution of "Approve" workflow event type. Without it, SPI cache in PAPI is out of date after workflow execution, which is cause issues described in JIRA issue.

### DIFF
--- a/kernel/classes/workflowtypes/event/ezapprove/ezapprovetype.php
+++ b/kernel/classes/workflowtypes/event/ezapprove/ezapprovetype.php
@@ -772,6 +772,12 @@ class eZApproveType extends eZWorkflowEventType
             $status = eZWorkflowType::STATUS_WORKFLOW_CANCELLED;
         }
         $contentObjectVersion->sync();
+
+        ezpEvent::getInstance()->notify(
+            'content/cache/version',
+            array( $contentObjectVersion->attribute( 'contentobject_id' ), $contentObjectVersion->attribute( 'version' ) )
+        );
+
         if ( $approvalStatus != eZApproveCollaborationHandler::STATUS_DEFERRED )
             $db->query( 'DELETE FROM ezapprove_items WHERE workflow_process_id = ' . $process->attribute( 'id' )  );
         return $status;


andrerom:
@gggeek This is the change we where talking about on slack, question was if we can solve your wait until date problem, the same way. Afaik there are no other place we really know if content where changed in workflow event types, so that is afaik why @adamwojs  ended up with this

andrerom:
@gggeek that sounds plausible, so the event is triggered from legacy cache clearing in eZContentCacheManager then most likely. But where is the transaction and the cache clear triggered?

gggeek:
The transaction is, I think, the default one guarding the legacy 'publish' operation.
It is that same operation that does indeed pass through `eZContentCacheManager::clearContentCacheIfNeeded ... eZContentCacheManager::clearNodeViewCacheArray`, which emits the event.
I am not able yet to explain why the behaviour is different when this happens via the cronjob rather than the backoffice.

andrerom:
Is the cronjob running via platform stack legacy script command?

gggeek:
yes

gggeek:
ps:
https://jira.ez.no/browse/EZP-28736
https://gist.github.com/gggeek/cfff2a602496c83dd30a70c7941e362b

andrerom:
@adamwojs given/if cache clear is actually sent, could you try alternative approach to this that might solve a lot more similar cases? This would avoid us having to pollute the whole code base with event calls.

- Add second optional array argument to `content/cache` for affected objectIds, and use it for clearing relevant content cache
- Same (also optional for BC) on ezcontentcachemanager::`clearNodeViewCacheArray()`, and pass in object id's from the two methods in same class calling it.
- not relevant to this issue, but as bonus do the same in eZContentOperationCollection::`removeNodes`

adamwojs:
Closed in favor of: 
* https://github.com/ezsystems/LegacyBridge/pull/140
* https://github.com/ezsystems/ezpublish-legacy/pull/1345

PR #1348, opened 2018-02-01
Update search index in hide cronjob

L0rdJ:


cronjobs/hide.php
```diff
@@ -49,6 +49,9 @@
                 {
                     $cli->output( 'Hiding node: "' . $node->attribute( 'name' ) . '" (' . $node->attribute( 'node_id' ) . ')' );
                     eZContentObjectTreeNode::hideSubTree( $node );
+                    
+                    //call appropriate method from search engine
```
andrerom: nitpick: `// call appropriate ..`

gggeek:
Seems legit.

I wonder if there are other cases in the legacy kernel where we miss similar calls...

eg: content/copy.php first creates a copy and only after it updates the visibility => if copy an hidden node, does the copy gets indexed as visible or invisible ?

andrerom:
@L0rdJ Seems some new commits have been introduced here now, possible to cleanup tp focus on the `cronjobs/hide.php` change so we can merge it? See also review comment.

L0rdJ:
@andrerom I pointed source branch for this PR.
I will check other possible cases where this call might be missed a bit later.

andrerom:
@L0rdJ Ok, but this PR is good to go right?

L0rdJ:
Yep, it safe to merge it to master as it is

extension/ezoe/design/standard/templates/ezoe/customattributes/link.tpl
```diff
@@ -90,6 +90,7 @@ eZOEPopupUtils.settings.onInitDoneArray.push( function( editorElement )
         ezoeLinkAttribute.timeOut = setTimeout( eZOEPopupUtils.BIND( ezoeLinkAttribute.ajaxCheck, this, url[0] + '_' + id, lid ), 320 );
         return true;
     });
+    inp.keyup();
```
andrerom: @L0rdJ I think you accidentally merge in unrelated changed here, or?

L0rdJ: @andrerom Sorry, you are correct. It was unrelated change. This change was related to https://github.com/ezsystems/ezpublish-legacy/pull/1291.

andrerom: Ok, merged that first so it got resolved when merging this, thanks @L0rdJ !

PR #1349, opened 2018-02-02
[WIP] Testing filter_input() on ezurloperator

glye:
> Work in progress

Can we make the `ezhttp` operator safer by default without breaking BC?

Should improve the test to show what is being stripped.
```
$_COOKIE['foo'] = '!b.&>;\/(#$A"snæfø</p>tag<snafu';
print( $_COOKIE['foo'] . "\n" );
print( filter_var($_COOKIE['foo'], FILTER_SANITIZE_STRING, FILTER_FLAG_STRIP_LOW|FILTER_FLAG_NO_ENCODE_QUOTES) . "\n" );
```
Results in
```
!b.&>;\/(#$A"snæfø</p>tag<snafu
!b.&>;\/(#$A"snæføtag
```

### DIFF
--- a/kernel/common/ezurloperator.php
+++ b/kernel/common/ezurloperator.php
@@ -796,7 +796,7 @@ CODEPIECE;
                         case eZURLOperator::HTTP_OPERATOR_TYPE_POST:
                         {
                             if ( $http->hasPostVariable( $httpName ) )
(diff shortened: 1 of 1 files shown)


glye:
After reading https://secure.php.net/manual/en/function.filter-input.php#99124 I'm trying `filter_var` instead.

PR #1350, opened 2018-02-06
Prevent fatal error

Jace25:
Under certain instances it can happen, that there is no contentobject with this id. In this case the setversionstatus would result in errors like 
`Fatal error: Call to a member function version() on null in /var/www/ezpublish_legacy/kernel/content/ezcontentoperationcollection.php on line 136`

Maybe some more work should be done here to check, why this function is called with a wrong objectId

### DIFF
--- a/kernel/content/ezcontentoperationcollection.php
+++ b/kernel/content/ezcontentoperationcollection.php
@@ -128,6 +128,9 @@ class eZContentOperationCollection
     static public function setVersionStatus( $objectID, $versionNum, $status )
     {
         $object = eZContentObject::fetch( $objectID );
+        if (!$object) {
+            return;
+        }
 
         if ( !$versionNum )
         {


andrerom:
Hi @Jace25,

yes as you say this is mainly silencing the messenger of a bug somewhere else. Do you perhaps have a trace or know anything more on the issue you are having? also feel free to reach out to ezcomunity slack to see if anyone might be able to help you find the bug.

Jace25:
Hi @andrerom,
i am trying to get more information for you. I got this when publishing stuff via legacykernelclosure. But only some publishings are failing, currently i have no glue why. I submit additional informations as soon as I have it.

But since the `$object = eZContentObject::fetch( $objectID );` can return null, i would suggest some check here, no matter where the error is triggered.

andrerom:
Ok, well could be db inconsistency.

Btw, any reason why you use legacy for this as opposed to Public API in Platform? Might give you a bit more solid experience, will also in most cases throw exceptions if something is wrong so it's possible to figure out what needs to be fixed.

Jace25:
Since we have lots of legacy workflows which should be triggered after publishing, i see no other way than using legacy publishing.

andrerom:
ok, that makes sense then.

For the future, be aware for after event things, you can potentially use signal slot system in Platform API to accomplish the same.

PR #1351, opened 2018-02-27
EZP-28881: Add a field to support "date object was trashed"

thiagocamposviana:
https://jira.ez.no/browse/EZP-28881

**Updated:** eZ Platform equivalent update: https://github.com/ezsystems/ezpublish-kernel/pull/2379

The eZ Publish trash can really fill and make it difficult to find something in the trash. It would be nice to be able to have a script that periodically clears out items that are older than "n days" (for example). In order to do this, we would need a field for objects in the trash that saves the "date object was trashed". It has been often requested by clients.

This pull request changes the eZContentObjectTrashNode so it stores the date the object has been trashed and then we can run the trashpurge script like:

php bin/php/trashpurge.php -s manage  --trashed-days=1

By default trashed days will be 0, so all content will be purged, if set, only objects that has been trashed for at least the specified amount of days will be purged.

For existing installations it is necessary to update the mysql table:

ALTER TABLE ezcontentobject_trash ADD trashed int(11) NOT NULL DEFAULT 0;

So, all existing trashed objects will have the trashed timestamp set to 0 and the script will be backcompatible.

SQL update has been added to https://github.com/ezsystems/ezpublish-kernel/pull/2261

### DIFF
--- a/kernel/classes/ezcontentobjecttrashnode.php
+++ b/kernel/classes/ezcontentobjecttrashnode.php
@@ -80,6 +80,10 @@ class eZContentObjectTrashNode extends eZContentObjectTreeNode
                                          'is_invisible' => array( 'name' => 'IsInvisible',
                                                                   'datatype' => 'integer',
                                                                   'default' => 0,
+                                                                  'required' => true ),
+                                         'trashed' => array( 'name' => 'Trashed',
+                                                                  'datatype' => 'integer',
+                                                                  'default' => 0,
                                                                   'required' => true )
                                           ),
 
@@ -130,7 +134,8 @@ class eZContentObjectTrashNode extends eZContentObjectTreeNode
                       'path_identification_string' => $node->attribute( 'path_identification_string' ),
                       'remote_id' => $node->attribute( 'remote_id' ),
                       'is_hidden' => $node->attribute( 'is_hidden' ),
-                      'is_invisible' => $node->attribute( 'is_invisible' ) );
+                      'is_invisible' => $node->attribute( 'is_invisible' ),
+                      'trashed' => time() );
 
         $trashNode = new eZContentObjectTrashNode( $row );
         return $trashNode;
@@ -207,6 +212,7 @@ class eZContentObjectTrashNode extends eZContentObjectTreeNode
                              'Limit'                    => false,
                              'SortBy'                   => false,
                              'AttributeFilter'          => false,
+                             'Trashed'                  => false,
                              );
         }
 
@@ -215,6 +221,7 @@ class eZContentObjectTrashNode extends eZContentObjectTreeNode
         $asObject         = ( isset( $params['AsObject']          ) )                         ? $params['AsObject']           : true;
         $objectNameFilter = ( isset( $params['ObjectNameFilter']  ) )                         ? $params['ObjectNameFilter']   : false;
         $sortBy           = ( isset( $params['SortBy']  ) && is_array( $params['SortBy']  ) ) ? $params['SortBy']              : array( array( 'name' ) );
+        $trashed          = ( isset( $params['Trashed']  ) && is_int( $params['Trashed'] )  ) ? " AND trashed <= {$params['Trashed']}"   : '';
 
         if ( $asCount )
         {
@@ -269,7 +276,8 @@ class eZContentObjectTrashNode extends eZContentObjectTreeNode
                         " . eZContentLanguage::sqlFilter( 'ezcontentobject_name', 'ezcontentobject' ) . "
                         $sqlPermissionChecking[where]
                         $objectNameFilterSQL
-                        AND " . eZContentLanguage::languagesSQLFilter( 'ezcontentobject' );
+                        AND " . eZContentLanguage::languagesSQLFilter( 'ezcontentobject' )
+                        . $trashed;
 
         if ( !$asCount && $sortingInfo['sortingFields'] && strlen( $sortingInfo['sortingFields'] ) > 5  )
             $query .= " ORDER BY $sortingInfo[sortingFields]";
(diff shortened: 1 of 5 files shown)


peterkeung:
+1

* Needs a Jira issue
* Yes, this should get an equivalent eZ Platform PR

kernel/classes/ezcontentobjecttrashnode.php
```diff
@@ -215,6 +221,7 @@ static function trashList( $params = false, $asCount = false )
         $asObject         = ( isset( $params['AsObject']          ) )                         ? $params['AsObject']           : true;
         $objectNameFilter = ( isset( $params['ObjectNameFilter']  ) )                         ? $params['ObjectNameFilter']   : false;
         $sortBy           = ( isset( $params['SortBy']  ) && is_array( $params['SortBy']  ) ) ? $params['SortBy']              : array( array( 'name' ) );
+        $trashed          = ( isset( $params['Trashed']  ) && is_int( $params['Trashed'] )  ) ? " AND trashed <= {$params['Trashed']}"   : '';
```
emodric: Should this be SQL escaped?

emodric:
I'm :+1: for this, but this definitelly needs the equivalent PR to eZ kernel, although this might be proven difficult to achieve.

This being a new feature, it will probably go into eZ kernel 7.1 or even 7.2. The problem is that legacy bridge by default always requires eZ legacy without upper limit, which means that new version which has this patch will happily install on eZ kernel 6.x too which will break.

thiagocamposviana:
@emodric 

Like this one https://github.com/ezsystems/ezpublish-kernel/pull/2261/files

emodric:
@thiagocamposviana Something like that, yes :)

kernel/private/classes/ezscripttrashpurge.php
```diff
@@ -106,7 +106,7 @@ public function run( $iterationLimit = 100, $sleep = 1 )
         }
         eZUser::setCurrentlyLoggedInUser( $user, $userCreatorID );
 
-        $trashCount = eZContentObjectTrashNode::trashListCount( false );
+        $trashCount = eZContentObjectTrashNode::trashListCount(  array( 'Trashed'  => $trashed ) );
         if ( !$this->quiet )
         {
             $this->cli->output( "Found $trashCount object(s) in trash." );
```
glye: This number may now be wrong. If `$trashed` is set, it should be amended to say that it's only counting objects beyond this age.

kernel/private/classes/ezscripttrashpurge.php
```diff
@@ -106,7 +106,7 @@ public function run( $iterationLimit = 100, $sleep = 1 )
         }
         eZUser::setCurrentlyLoggedInUser( $user, $userCreatorID );
 
-        $trashCount = eZContentObjectTrashNode::trashListCount( false );
+        $trashCount = eZContentObjectTrashNode::trashListCount(  array( 'Trashed'  => $trashed ) );
```
glye: CS nitpick: Extra space before `=>`.

kernel/private/classes/ezscripttrashpurge.php
```diff
@@ -123,7 +123,7 @@ public function run( $iterationLimit = 100, $sleep = 1 )
         while ( $trashCount > 0 )
         {
             $this->monitor( "iteration start" );
-            $trashList = eZContentObjectTrashNode::trashList( array( 'Limit'  => $iterationLimit ), false );
+            $trashList = eZContentObjectTrashNode::trashList( array( 'Limit'  => $iterationLimit, 'Trashed'  => $trashed ), false );
```
glye: CS nitpick: Extra space before `=>`.

kernel/private/classes/ezscripttrashpurge.php
```diff
@@ -143,7 +143,7 @@ public function run( $iterationLimit = 100, $sleep = 1 )
                 return false;
             }
 
-            $trashCount = eZContentObjectTrashNode::trashListCount( false );
+            $trashCount = eZContentObjectTrashNode::trashListCount( array( 'Trashed'  => $trashed ) );
```
glye: CS nitpick: Extra space before `=>`.

kernel/private/classes/ezscripttrashpurge.php
```diff
@@ -72,7 +72,7 @@ public function __construct( eZCLI $cli, $quiet = true, $memoryMonitoring = fals
      *
```
glye: Missing a `@param` description of the new `$trashed` parameter here.

thiagocamposviana:
Updated the code.

thiagocamposviana:
The sql errors are probably due to the sql changes being made for the ez platform stack now? If the ez platform pull request passes we should trigger a new build for this one?

andrerom:
Given the PR on kernel was closed, should we also close this one? No need for the feature anymore?

thiagocamposviana:
We do need the feature in both, but the way it is being handled I don't know if I can keep provide support for it any further, I will need to talk internally to discuss a viable strategy to submit this kind of requests.

andrerom:
ok

emodric:
@andrerom What about support for this in the new kernel? Is this BC compatible with the new kernel?

andrerom:
This one?
https://github.com/ezsystems/ezpublish-kernel/pull/2379

ref why conflict was bundled once both was merged https://github.com/ezsystems/ezpublish-legacy/commit/db9e6b10c7af96469ca3ecf29e8b2261798125ad

emodric:
Ah, okay, I missed that it was merged. "Closed" status threw me off.

:+1: about conflict!

andrerom:
indeed, unsure if there is a way to tell gh to mark it as merged and not closed 🤷‍♂️

andrerom:
Follow up on missing SQL updates / schema changes as preparations for release: https://github.com/ezsystems/ezpublish-legacy/pull/1393

PR #1352, opened 2018-03-02
Prevent ImagealiasHandler from storing invalid chars

Jace25:
Some exif data may contain invalid chars which result in fatal errors when trying to update this content. This PR removes this chars and fixes this issue.

See as example the char at COMPUTED['UserComment']
image

Without this fix `\eZ\Publish\Core\Persistence\Legacy\Content\FieldValue\Converter\ImageConverter::parseLegacyXml`would run into this fatal error:

> Fatal error: Call to a member function hasAttribute() on null in /var/www/ez/source/ez-publish/vendor/ezsystems/ezpublish-kernel/eZ/Publish/Core/Persistence/Legacy/Content/FieldValue/Converter/ImageConverter.php on line 189

### DIFF
--- a/kernel/classes/datatypes/ezimage/ezimagealiashandler.php
+++ b/kernel/classes/datatypes/ezimage/ezimagealiashandler.php
@@ -1451,6 +1451,13 @@ class eZImageAliasHandler
         $this->ContentObjectAttributeData['DataTypeCustom']['dom_tree'] = $domTree;
         $this->ContentObjectAttributeData['DataTypeCustom']['is_storage_required'] = false;
         $xmlString = $domTree->saveXML();
+
+        // Remove chars that are invalid for XML 1.0, see https://www.w3.org/TR/2008/REC-xml-20081126/#charsets
+        $xmlString = preg_replace(
+            '/[^\x{0009}\x{000A}\x{000D}\x{0020}-\x{D7FF}\x{E000}-\x{FFFD}\x{10000}-\x{10FFFF}]+/u', 
+            '', $xmlString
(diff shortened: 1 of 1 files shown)


gggeek:
The end goal is good. However I wonder if this is the best place to do this... Do you know what is the source of those invalid bytes ?

Jace25:
I am not 100% sure. I think some constellation of user error using windows together with some exif editor tools. Often this are images delivered by press agencys like dpa, but also from our own editors.

here some exiftool output, where you can see the invisible char:
`ExifTool Version Number         : 8.60
File Name                       : Anton.jpg
Directory                       : .
File Size                       : 333 kB
File Modification Date/Time     : 2018:03:02 13:58:32+01:00
File Permissions                : rw-rw-r--
File Type                       : JPEG
MIME Type                       : image/jpeg
JFIF Version                    : 1.01
Exif Byte Order                 : Big-endian (Motorola, MM)
Processing Software             : Windows Photo Editor 10.0.10011.16384
Make                            : SAMSUNG
Camera Model Name               : GT-I9300
...
Software                        : Windows Photo Editor 10.0.10011.16384
Modify Date                     : 2018:02:24 13:20:20
...
User Comment                    : .�.;.
...
Compression                     : JPEG (old-style)
...
Creator Tool                    : Windows Photo Editor 10.0.10011.16384
Image Width                     : 1353
Image Height                    : 1015
...`

Notice the User Comment

Jace25:
Any updates here?

Jace25:
thx for nothing

PR #1355, opened 2018-03-27
EZP-29033: [Legacy] Don't remove links user has no access to

glye:
> Fix https://jira.ez.no/browse/EZP-29033
> QA

The security fix EZSA-2018-001 ensured that the names of protected content could not be leaked by making content/node links to it. The links would normally be rendered as URL aliases, which leaks the content name. After the fix, links are rendered empty, making them non-functional.

This fixes it: When the user doesn't have read access to the linked node/object, show a `/content/view/full/<nodeId>` link rather than nothing. So the link is functional, the user is told they don't have access, while not leaking the content name. This is the new default behaviour. You can set `ezxml.ini [ezxhtml] ShowURLAliasForProtectedLinks=enabled` to get the old, pre-security-fix behaviour (always URL alias).

This affects only eznode/ezobject links in XML text, not related objects.


glye:
Update: Added setting `ezxml.ini [ezxhtml] ShowURLAliasForProtectedLinks` to control this behaviour.

glye:
Thanks! Sent to QA.

PR #1356, opened 2018-03-27
EZP-29085: Force to use integer keys in eZXMLOutputHandler->NodeArray

L0rdJ:
> Fix https://jira.ez.no/browse/EZP-29085
> Sent to QA

Copied from Slack:
Seems like nodeID is forced to convert to string at https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/classes/datatypes/ezxmltext/ezxmloutputhandler.php#L278 and https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/classes/datatypes/ezxmltext/ezxmloutputhandler.php#L285

But http://php.net/manual/en/language.types.array.php says:
_Additionally the following key casts will occur:
Strings containing valid decimal integers, unless the number is preceded by a + sign, will be cast to the integer type. E.g. the key “8” will actually be stored under 8. On the other hand “08" will not be cast, as it isn’t a valid decimal integer._

I.e. the string conversion is pointless, the array index will be an integer anyway.

### DIFF
--- a/kernel/classes/datatypes/ezxmltext/ezxmloutputhandler.php
+++ b/kernel/classes/datatypes/ezxmltext/ezxmloutputhandler.php
@@ -275,14 +275,14 @@ class eZXMLOutputHandler
                 foreach( $nodes as $node )
                 {
                     $nodeID = $node->attribute( 'node_id' );
-                    $this->NodeArray["$nodeID"] = $node;
+                    $this->NodeArray[$nodeID] = $node;
                 }
             }
             elseif ( $nodes )
             {
                 $node = $nodes;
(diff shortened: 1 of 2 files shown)


L0rdJ:
More details discussed on Slack with Gunnstein Lye

glye:
@L0rdJ Could you please make the commit text more descriptive? And if the two fixes are for separate issues, can you put them in separate commits, please?

L0rdJ:
@glye i changed PR title and commit message. That two fixes are not separate, so they should be in the same commit.

glye:
Sent to QA.

PR #1358, opened 2018-04-04
EZP-29039: Image aliases expiry on draft discard

kmadejski:
JIRA issue: https://jira.ez.no/browse/EZP-29039

## Description

As described in JIRA issue, the image aliases are always set as expired in DFS `ezdfsfile` table during draft rejection under two conditions existing simultaneously:
* Image parent node (folder/gallery) has to has special characters in the name/title (for instance quotes)
* `[URLTranslator]` setting needs to be set as `TransformationGroup=urlalias_iri`

Due to `TransformationGroup=urlalias_iri` setting names of directories on the filesystem also contain special characters.

An image is saved to the `ezcontentobject_attribute`.`data_text` as XML. Each XML node which represents image/image alias has attribute `url`. Under the mentioned conditions, during the conversion phase, the value of this attribute is created from the file path and all special characters are converted to HTML entities (in this case `"` to `&quot;`). 

During draft discarding, `eZImageFile::isReferencedByOtherAttributes` method is invoked. All special characters in the XML `url` attribute value mentioned before should be also converted to the HTML entities. Otherwise, that method returns `0` what means, that there are no more references and the file can be safely removed.

### DIFF
--- a/kernel/classes/datatypes/ezimage/ezimagefile.php
+++ b/kernel/classes/datatypes/ezimage/ezimagefile.php
@@ -248,7 +248,7 @@ class eZImageFile extends eZPersistentObject
                 AND ezcontentobject_attribute.id = %d
                 AND ( ezcontentobject_attribute.version != %d OR ezcontentobject_attribute.language_code != '%s' )
                 ",
-                $filepath, $filepath,
+                $filepath, htmlspecialchars($filepath),
                 $attributeId,
                 $attributeVersion,
                 $languageCode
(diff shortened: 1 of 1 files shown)


kmadejski:
@gggeek thank you for the feedback. Good catch. 

> htmlspecialchars also used explicitly when saving the data in data_text, or is the escaping happening by some xml-conversion process?

This is done by pure XML conversion when `DOMElement::setAttribute` is executed. I did some test and changed `htmlspecialchars()` flags to `ENT_XML1 | ENT_COMPAT`. During the test which I did, I didn't notice any problem with this approach.

andrerom:
@gggeek Looks ok to you now?

kmadejski:
@glye thanks for the review. Yes, I've checked also `ezfile/ezbinaryfile` but these are stored in DFS like: `var/ezdemo_site/storage/original/video/280c6a635fd32a4df6f18baa2601f93b.mp4` so, not with their real names.

kmadejski:
@gggeek could you please take a glance?

kmadejski:
@gggeek may I ask you for the final review, please?

PR #1359, opened 2018-04-06
FIX cache not cleared on node visibility change

ITernovtsiy:
Fix issue with cache not cleared on node visibility change.
Search index takes wrong visibility status from cache.

Todo:
- [x] 2017.12
- [x] JIRA issue - https://jira.ez.no/browse/EZP-30493

### DIFF
--- a/kernel/content/ezcontentoperationcollection.php
+++ b/kernel/content/ezcontentoperationcollection.php
@@ -1014,7 +1014,9 @@ class eZContentOperationCollection
                 $action = 'show';
             }
             else
+            {
                 eZContentObjectTreeNode::hideSubTree( $curNode );
+            }
 
             // Notify cache system about visibility change
             ezpEvent::getInstance()->notify('content/cache', [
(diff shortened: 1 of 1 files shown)


kernel/content/ezcontentoperationcollection.php
```diff
@@ -1017,6 +1017,12 @@ static public function changeHideStatus( $nodeID )
                 eZContentObjectTreeNode::hideSubTree( $curNode );
         }
 
+        // notify cache system about visibility change
+        ezpEvent::getInstance()->notify('content/cache', array(
```
andrerom: maybe move this inside ` if ( is_object( $curNode ) )`? otherwise nothing is actually changed

andrerom:
Updated PR here @diaborn19, would you be able to move the changes either with the two commits or just one by you to 2017.12 branch?

ITernovtsiy:
Hi @andrerom 
Sorry, missed your comment, changed to 2018.09

andrerom:
I meant 2017.12 here, it's the release used with eZ Platform 1.13.x, so the oldest branch here that receives fixes for the time being. But if you prefer I can adapt this on merge.

ITernovtsiy:
Hi, EZP-30493 created and base branch switched to 2017.12. Thanks.

PR #1360, opened 2018-05-19
Avoid double decoding of URL

pkamps:
eZSYS::requestURI is already decoding the URL. The eZURI class does it a 2nd time.

Example of the problem. Given URL:
http://<sitedomain>/(foo)/bar%2Bbar

Double decoded version is is /(foo)/bar bar (that's a space in between)
But the correct value is /(foo)/bar+bar

You can test this by adding following into your pagelayout.tpl:
{$view_parameters|dump()}

Use following URL and look at the dump output:
http://<sitedomain>/(foo)/bar%2Bbar

The correct value is " /(foo)/bar+bar" - the pull request should produce the correct output.

### DIFF
--- a/lib/ezutils/classes/ezuri.php
+++ b/lib/ezutils/classes/ezuri.php
@@ -576,7 +576,10 @@ class eZURI
         {
             if ( !isset( $GLOBALS['eZURIRequestInstance'] ) )
             {
-                $GLOBALS['eZURIRequestInstance'] = new eZURI( eZSys::requestURI() );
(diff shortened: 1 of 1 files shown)


pkamps:
Issue:  https://jira.ez.no/browse/EZP-29246

pbek:
@pkamps, FYI: I tested your pull request and it didn't solve the issue of showing customer information with urls like <https://your-server/admin/shop/customerorderview/10/test%2Btest@test.com>

PR #1361, opened 2018-05-23
EZP-28950: Add support for utf8mb4 charset to i18n

emodric:
> https://jira.ez.no/browse/EZP-28950

eZ legacy does not recognize `utf8mb4` charset, instead falling back to conversion to a single byte charset set by `eZTextCodec`, which results in the following beautiful error messages:

```
Query error (1064): You have an error in your SQL syntax; check the manual that
corresponds to your MySQL server version for the right syntax to use near '???????????????????????????????????????????????????????????
?????????????????????' at line 1. Query: ?????????????????????????????????????????????????????????????????
??????????????????????????????????????????????????????????????????????
[ May 23 2018 10:47:03 ] Unexpected error, the message was : Call to a member function
attribute() on null in vendor/ezsystems/legacy-bridge/bundle/LegacyMapper/Security.php on line 97
```

The effect of this PR is that `utf8mb4` is internally recognized as `utf-8`, resulting in no conversion from UTF8 taking place when used for example by the database layer.

Disclaimer: I'm not claiming that this is a proper fix, especially in this obscure part of legacy, but it seems logical :)

@glye @andrerom @gggeek @alongosz Any input on your side?

### DIFF
--- a/lib/ezi18n/classes/ezcharsetinfo.php
+++ b/lib/ezi18n/classes/ezcharsetinfo.php
@@ -150,7 +150,7 @@ class eZCharsetInfo
                                                            'cp849',
                                                            'cp850' ),
                                     'unicode' => array( 'unicode' ),
-                                    'utf-8' => array( 'utf-8' ) );
+                                    'utf-8' => array( 'utf-8', 'utf8mb4' ) );
         }
         return $encodingTable;
     }


glye:
Note: This is a prerequisite for `utf8mb4` to work on legacy, but doesn't stop the old `utf-8` from working. Should be backported to all supported branches.

gggeek:
+1

glye:
For reference, before this PR, legacy doesn't know what `utf8mb4` is, and assumes it is a singlebyte encoding, which obviously fails. See https://github.com/ezsystems/ezpublish-legacy/blob/297ff7cbf1f976a0f544fc6c2c0c111ca2743a80/lib/ezi18n/classes/ezcharsetinfo.php#L213

After the fix, this returns `utf-8`, which means that input and output is the same, and eZTextCodec does nothing to our SQL strings, which is what we want. `utf-8` in PHP is fully compatible with `utf8mb4` in MySQL. See https://github.com/ezsystems/ezpublish-legacy/blob/297ff7cbf1f976a0f544fc6c2c0c111ca2743a80/lib/ezdb/classes/ezmysqlidb.php#L271

glye:
Thanks @emodric & @micszo! Merged.

PR #1364, opened 2018-05-29
Fix admin UI anchor link scroll bug

benkmugo:
- fixed height on the width controller element causes the user to get
  'stuck' and unable to scroll back to the top when clicking on a link
targeting an in page anchor

The bug is also referenced in this Jira ticket:
https://jira.ez.no/browse/EZP-29210


benkmugo:
@glye thanks for the quick response Gunnstein.
An issue with the change affecting the draggable sidebar area was discovered last night, which wasn't found in testing previously.

I'll close this PR and will open a new one once that issue has been fixed and a fresh round of testing is done on our end.

PR #1366, opened 2018-06-01
EZP-29210: Fix admin UI anchor link scroll bug

benkmugo:
- fixed height on the width controller element causes the user to get
'stuck' and unable to scroll back to the top when clicking on a link
targeting an in page anchor

The bug is also noted in this Jira ticket:
https://jira.ez.no/browse/EZP-29210

### DIFF
--- a/design/admin/stylesheets/pagelayout.css
+++ b/design/admin/stylesheets/pagelayout.css
@@ -472,11 +472,10 @@ ul.leftmenu-items li.current a
 #widthcontrol-handler
 {
     position: absolute;
-    right: -30px;
-    top:-20px;
+    right: 20px;
+    top: 0;
     bottom: 0; /* strech div to full width of relative box */
     width: 10px;
-    height: 5000px;
     cursor: col-resize;
     z-index:11;
 /*    background: #f8f8f8;*/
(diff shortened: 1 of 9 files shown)


benkmugo:
This new PR fixes the original issue and addresses the side-effect found on PR #1364 (closed)

emodric:
I've reproduced the issue, so if this fixes it, I'm +1. I do however worry about 3rd party extensions that might use `<div id="widthcontrol-handler">` in their own left menu implementations. What happens where there are two such elements on the page, one from the extension and one from pagelayout?

konradoboza:
This fix works (checked on Chrome, Firefox and Safari). However, is there any reason why `web/design/admin/templates/parts/user/menu.tpl` wasn't also taken into account?

benkmugo:
@emodric the instance I was testing this on had the eZ Tags extension installed, which also has a menu include with the widthcontrol-handler in it. When there were multiple widthcontrol-handler elements in place (one from the pagelayout, one from the menu include) only the first one from the pagelayout was shown and active.
The others remained hidden/inactive. You can double-check, by adding back that element in one of the menu template includes to verify.

@konradoboza it looks like I missed that one. I'll check and update the commit shortly.

Thanks all for the quick feedback.

benkmugo:
The update has been applied to the user template as well now, as pointed out by @konradoboza above.

PR #1367, opened 2018-06-05
Fixed fatal error

omh:
`continue` used outside of loop.


andrerom:
@omh Could you create jira issues for the PR's you have done the last month?

omh:
Issue: https://jira.ez.no/browse/EZP-29351

andrerom:
>  Do you want me to get my colleagues (Rune, Jørgen, Elise) to do the same?

For anything affecting prior releases, yes ideally we would want that for change logs.

PR #1368, opened 2018-06-05
EZP-29350: Make query compatible with strict MySQL

omh:
MySQL 5.7 is strict by default and will error out on the incorrect GROUP BY in the update script (see diff).


omh:
Issue: https://jira.ez.no/browse/EZP-29350

PR #1369, opened 2018-06-05
EZP-29285: Custom tags: CTRL+Enter to get between 2 consecutive tags

thiagocamposviana:
https://jira.ez.no/browse/EZP-29285

if you have 2 consecutive custom tags in an XML block, you cannot insert text between then unless you disable the editor.


thiagocamposviana:
Yes, we use it in some projects.

emodric:
If it works, I wouldn't mind having it enabled by default :)

andrerom:
@thiagocamposviana Ok with you?

thiagocamposviana:
thanks!

PR #1370, opened 2018-06-25
Fixed php7 warning

elisekrist:
ErrorException (E_WARNING)
Illegal string offset 'path'

### DIFF
--- a/kernel/content/edit.php
+++ b/kernel/content/edit.php
@@ -765,6 +765,7 @@ if ( !function_exists( 'checkContentActions' ) )
                         }
                         else if ( isset( $operationResult['result'] ) )
                         {
+                            $Result = array();
                             $result = $operationResult['result'];
                             $resultContent = false;
                             if ( is_array( $result ) )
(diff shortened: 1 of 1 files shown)


kernel/content/edit.php
```diff
@@ -775,7 +775,7 @@ function computeRedirect( $module, $object, $version, $EditLanguage = false )
                                 }
                                 if ( isset( $result['path'] ) )
                                 {
-                                    $Result['path'] = $result['path'];
+                                    $Result = array('path' => $result['path']);
```
andrerom: It's potentially similar issue below.

Safer and more readable would be to set $Result = array(); at the start of `else if ( isset( $operationResult['result'] ) )` block.

andrerom:
Besides that nice catch, and welcome :)

kernel/content/edit.php
elisekrist: Thanks for the feedback. I moved $Result to the top of the case.

PR #1371, opened 2018-06-27
EZP-27101: Recursion protection for retrieving metadata doesn't work properly

natanael89:
JIRA issue: [EZP-27101](https://jira.ez.no/browse/EZP-27101)

When retrieving the search metadata from objects containing "Object relation" and "Object relations" field type, there is a recursion protection in place that ensures that there is no infinite recursion caused by circular references. However, it only checks if the certain object was already processed and this is causing problems when the same object is referenced from two different fields (as described in [EZP-27101](https://jira.ez.no/browse/EZP-27101)). To fix that I introduced new recursion protection that differs in two aspects:
* Besides object id, it also stores the field id which contains this object. Circular reference only happens if we are processing the same object from the same field as before. 
* It works like a stack instead of a one-dimensional array, meaning that only references above in the recursion chain are taken into the account when detecting circular references.

There is another PR which tries to fix this issue (https://github.com/ezsystems/ezpublish-legacy/pull/1288), but it can cause an infinite recursion (as described in https://github.com/ezsystems/ezpublish-legacy/pull/1288#issuecomment-292988830).

### DIFF
--- a/kernel/classes/datatypes/ezobjectrelation/ezobjectrelationtype.php
+++ b/kernel/classes/datatypes/ezobjectrelation/ezobjectrelationtype.php
@@ -639,7 +639,7 @@ class eZObjectRelationType extends eZDataType
         $object = $this->objectAttributeContent( $contentObjectAttribute );
         if ( $object )
         {
-            if ( eZContentObject::recursionProtect( $object->attribute( 'id' ) ) )
+            if ( eZContentObject::stackRecursionProtect( $contentObjectAttribute->attribute( 'id' ), $object->attribute( 'id' ) ) )
             {
                 // Does the related object exist in the same language as the current content attribute ?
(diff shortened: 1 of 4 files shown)


pkamps:
We also had issues with the way the metaData was retrieved. Not only technical because the logic was fetching unrelated data which is resulting in bad search results.
We fixed the issue in this pull request:
https://github.com/mugoweb/ezpublish-legacy/pull/55

natanael89:
@pkamps Yes, I saw your PR. In general I agree that the search retrieves too much information for most use cases, but I'm not sure if we can change the logic now since that would be a BC break.

pkamps:
The recursive lookup is currently technically broken and we agree it's a bad/broken concept. Not sure why you want to keep it. It's unrealistic that somebody relies on a broken "feature".

natanael89:
After the internal discussion, we decided to close this PR. I'll make another one which will remove the recursive logic in the least BC-breaky way as possible.

PR #1372, opened 2018-06-28
EZP-27101: Remove recursion when retrieving search metadata for relation fields

natanael89:
JIRA issue: [EZP-27101](https://jira.ez.no/browse/EZP-27101)

When retrieving the search metadata from objects containing "Object relation" and "Object relations" field type, there is a recursion protection in place that ensures that there is no infinite recursion caused by circular references. However, it only checks if the certain object was already processed and this is causing problems when the same object is referenced from two different fields (as described in [EZP-27101](https://jira.ez.no/browse/EZP-27101)). 

This PR removes the recursion completely by making the relations load metadata only one level deep. By removing recursion, I was also able to remove the recursion protection, which solves the original issue.

This PR introduces a slight BC break because it changes logic (previously the metadata was loaded recursively and now it is loaded only one level deep), but the new behavior allows for better (more relevant) search results.
An alternative solution (without a BC break, but which keeps the broken logic of loading the metadata recursively) can be found here: https://github.com/ezsystems/ezpublish-legacy/pull/1371.

### DIFF
--- a/doc/bc/5.90/relation_indexing.md
+++ b/doc/bc/5.90/relation_indexing.md
@@ -1,0 +1,13 @@
+# Relation search indexing changes 
+
+For the 2018.06 release, eZ Publish changed the way relations are indexed. Previously the relation metadata was loaded recursively, meaning that relations of relations etc. were also included. This behavior was changed to only include Objects one level deep so only direct relations are taken into account.
+
+The affected kernel Datatypes are eZObjectRelationType and eZObjectRelationListType.
+
+If you have your custom Datatype that stores relations to other Objects, you can make it behave similarly. Just add the following to your Datatype class:
+```php
+    public function isRelationType()
+    {
+        return true;
+    }
+```
--- a/doc/bc/5.90/relation_indexing.md
+++ b/doc/bc/5.90/relation_indexing.md
@@ -2,6 +2,8 @@
 
 For the 2018.06 release, eZ Publish changed the way relations are indexed. Previously the relation metadata was loaded recursively, meaning that relations of relations etc. were also included. This behavior was changed to only include Objects one level deep so only direct relations are taken into account.
 
+The change was done to avoid various recursion problems when relations are circular. Also, the new behavior allows for better (more relevant) search results.
+
 The affected kernel Datatypes are eZObjectRelationType and eZObjectRelationListType.
 
 If you have your custom Datatype that stores relations to other Objects, you can make it behave similarly. Just add the following to your Datatype class:
--- a/kernel/classes/ezcontentobjectattribute.php
+++ b/kernel/classes/ezcontentobjectattribute.php
@@ -1182,7 +1182,7 @@ class eZContentObjectAttribute extends eZPersistentObject
             }
 
             $datatype = $attribute->dataType();
-            if( $skipRelationAttributes && $datatype->isRelationType() )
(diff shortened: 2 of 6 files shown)


andrerom:
@natanael89 for the BC break, can you add description of that in doc/bc/5.90/relation_indexing.md ?

This will be for the 2018.06 release.

kernel/classes/ezcontentobjectattribute.php
```diff
@@ -1168,26 +1168,34 @@ function fromString( $string )
...
-            if ( $classAttribute->attribute( 'is_searchable' ) )
+            if ( !$classAttribute->attribute( 'is_searchable' ) )
             {
-                $attributeMetaData = $attribute->metaData();
-                if ( $attributeMetaData !== false )
+                continue;
+            }
+
+            $datatype = $attribute->dataType();
+            if($skipRelationAttributes && $datatype->isRelationType())
```
andrerom: legacy cs: `if( $skipRelationAttributes && $datatype->isRelationType() )`

kernel/search/plugins/ezsearchengine/ezsearchengine.php
```diff
@@ -68,7 +68,6 @@ public function addObject( $contentObject, $commit = true )
         $placement = 0;
         $previousWord = '';
 
-        eZContentObject::recursionProtectionStart();
```
andrerom: wondering if we might need to keep these in, in case custom relation field types stiull use it.

or will it cause problems?

natanael89: We can (and should) keep those, you are right.

andrerom:
@pkamps Is this more in line with what you suggested i other other PR? Do you approve? :)

@gggeek Any opinions on this in terms of search use?

natanael89:
I added the BC description and made the suggested fixes.

kernel/classes/ezcontentobjectattribute.php
```diff
@@ -1182,7 +1182,7 @@ static function metaDataArray( &$attributes, $skipRelationAttributes = false )
             }
 
             $datatype = $attribute->dataType();
-            if($skipRelationAttributes && $datatype->isRelationType())
+            if( $skipRelationAttributes && $datatype->isRelationType() )
```
andrerom: my prev comment forgot to also specify whitespace after `if`

doc/bc/5.90/relation_indexing.md
```diff
@@ -0,0 +1,13 @@
+# Relation search indexing changes 
+
+For the 2018.06 release, eZ Publish changed the way relations are indexed. Previously the relation metadata was loaded recursively, meaning that relations of relations etc. were also included. This behavior was changed to only include Objects one level deep so only direct relations are taken into account.
```
andrerom: should explain why the change was done.

like `are taken into account, in order to solve various recursion problems when relations are circular.`

kernel/classes/ezcontentobjectattribute.php
natanael89: Fixed.

doc/bc/5.90/relation_indexing.md
natanael89: I added another sentence for the "why" part.

PR #1373, opened 2018-07-06
EZP-29390: [Legacy] PHP Warning: Cannot change session name when session is active

kmadejski:
| Question           | Answer
| **JIRA issue**     | [EZP-29390](https://jira.ez.no/browse/EZP-29390)
| **Bug**| yes
| **New feature**    | no
| **Target version** | `master` 

PHP 7.2 emits warning `PHP Warning:  session_name(): Cannot change session name when session is active`. In the previous versions, it was not possible to change session name when a session already exists too, but there was no warning.

More info: http://php.net/manual/en/function.session-name.php#refsect1-function.session-name-changelog & https://stackoverflow.com/questions/47700336/php-7-2-warning-cannot-change-session-name-when-session-is-active

### DIFF
--- a/lib/ezsession/classes/ezsession.php
+++ b/lib/ezsession/classes/ezsession.php
@@ -262,6 +262,9 @@ class eZSession
      *
      * @since 4.1
      * @return bool Depending on if eZSession is registrated as session handler.
+     *
+     * @since v2018.06.0
+     * @return false if session is already active since PHP 7.2 onwards emits a warning about this.
     */
     static protected function registerFunctions( $sessionName = false, ezpSessionHandler $handler = null )
     {
--- a/lib/ezsession/classes/ezsession.php
+++ b/lib/ezsession/classes/ezsession.php
@@ -261,16 +261,15 @@ class eZSession
      * {@link eZSession::start()}, so only call this if you don't start the session.
      *
      * @since 4.1
-     * @since v2018.06.0 Return false if session is already active since PHP 7.2 onwards emits a warning about this.
      * @return bool Depending on if eZSession is registrated as session handler.
     */
     static protected function registerFunctions( $sessionName = false, ezpSessionHandler $handler = null )
     {
-        if ( self::$hasStarted || self::$handlerInstance !== null || session_status() === PHP_SESSION_ACTIVE )
+        if ( self::$hasStarted || self::$handlerInstance !== null )
             return false;
 
         $ini = eZINI::instance();
-        if ( $sessionName !== false )
+        if ( $sessionName !== false && $sessionName !== session_name() )
         {
             session_name( $sessionName );
         }
@@ -283,7 +282,9 @@ class eZSession
                 // Use md5 to make sure name is only consistent of alphanumeric characters
                 $sessionName .=  md5( $access['name'] );
             }
-            session_name( $sessionName );
+            if ( $sessionName !== session_name() ) {
+                session_name( $sessionName );
+            }
         }
         else
         {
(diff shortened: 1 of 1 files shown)


kmadejski:
Tests didn't pass because of external service (http://soap.critmon1.ez.no) which is not available at the moment - it returns 503 error code. I'll try to restart tests after some time.

andrerom:
Is this called only to set session name, or is it also doing more stuff which you now block?

andrerom:
Thinking especially about:
https://github.com/kmadejski/ezpublish-legacy/blob/5c2430feb6165e83b0e1b9c4ae2786bec88ff380/lib/ezsession/classes/ezsession.php#L293-L308

kmadejski:
@andrerom it's pretty tricky to predict all cases where it might be used. Therefore, I've assumed that you have right and this might be problematic in some cases, so I've changed the approach here to another, more bulletproof one in https://github.com/ezsystems/ezpublish-legacy/pull/1373/commits/dff1f02b4ae7a06de3b081e5a288a7d4bdc6c1f7.

lib/ezsession/classes/ezsession.php
```diff
@@ -261,16 +261,15 @@ static public function countActive()
...
     */
     static protected function registerFunctions( $sessionName = false, ezpSessionHandler $handler = null )
     {
-        if ( self::$hasStarted || self::$handlerInstance !== null || session_status() === PHP_SESSION_ACTIVE )
+        if ( self::$hasStarted || self::$handlerInstance !== null )
             return false;
 
         $ini = eZINI::instance();
-        if ( $sessionName !== false )
+        if ( $sessionName !== false && $sessionName !== session_name() )
```
andrerom: Alternative is to move all this session name logic into own method and only call it if session_status() !== PHP_SESSION_ACTIVE

But that said I'm ok with this, also, smaller diff.

andrerom:
@glye your review still stands?

glye:
Not quite sure how the current approach works. If the session is started, the supplied name will always match the current name, so renaming can be skipped, avoiding the warning? If that's the idea and you're confident of it, I'm cool.

kmadejski:
> If the session is started, the supplied name will always match the current name, so renaming can be skipped, avoiding the warning?

Yes, exactly, this is the idea. I checked it even on full legacy mode and everything seems to be fine. T

andrerom:
@kmadejski I think I fixed your comment, you mean you tried out running pure legacy right?
ref: https://doc.ez.no/display/EZP/Architecture#Architecture-SummaryonthewaystouseeZPublish5

lib/ezsession/classes/ezsession.php
pkamps: Good point.

emodric:
Hi! After this, when running legacy scripts through Symfony console command, PHP emits a notice:

```
eddie@abyss [~/www/netgensite] ( ± master ● ) $ php bin/console ezpublish:legacy:script bin/php/ezpgenerateautoloads.php
PHP Notice:  Undefined index: eZCurrentAccess in /media/eddie/data/html/netgensite/ezpublish_legacy/lib/ezsession/classes/ezsession.php on line 281
Scanning for PHP-files.

Scan complete. Found 526 PHP files.

Found 207 classes, added 207 of them to the autoload array.

eddie@abyss [~/www/netgensite] ( ± master ● ) $ 
```

blankse:
Hi, after this (Update from v2017.12.3 to v2017.12.3.1) I have following problem:
When I login in the backend. I have rights in the legacy but in the new stack.

if I load content over the new stack I am not the logged in user but the anonymous user...

kmadejski:
@emodric, @blankse thanks for reporting it. I'll take a look at it as soon as possible.

PR #1374, opened 2018-07-06
Skip tests that depend on critmon1.ez.no as it has been shut down

glye:
Whoops, forgot about these old tests when killing critmon.
As discovered in https://github.com/ezsystems/ezpublish-legacy/pull/1373


glye:
@andrerom Yes, will look into making a replacement as we don't have one afaik. Likely a simple vhost on support.ez.no.

PR #1375, opened 2018-07-08
Purge preview cache in `attribute_edit.php`

zerustech:
The admin preview function does not work as expected: it always fetches
content from the view cache, so the draft content being edited never
shows up.

To fix this issue, view cache should be purged by triggering listeners
attached to `content/cache/translations`.

This PR is related to ezsystems/LegacyBridge#153

| Question           | Answer
| **JIRA issue**     | n/a
| **Bug**            | yes
| **New feature**    | no
| **Target version** | `2017.12`
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | no

**TODO**:
- [ ] Implement feature / fix a bug.
- [ ] Implement tests.
- [ ] Fix new code according to Coding Standards (`$ composer fix-cs`).
- [x] Ask for Code Review.

### DIFF
--- a/kernel/content/attribute_edit.php
+++ b/kernel/content/attribute_edit.php
@@ -325,6 +325,11 @@ if ( $storingAllowed && $hasObjectInput)
             'content/cache/version',
             array( $object->attribute( 'id' ), $version->attribute( 'version' ) )
         );
+
+        ezpEvent::getInstance()->notify(
+            'content/cache/translations',
+            array( $object->attribute( 'id' ), $version->attribute( 'version' ) , [$EditLanguage])
+        );
     }
 
     $validation['processed'] = true;


kernel/content/versionview.php
```diff
@@ -13,6 +13,8 @@
 $LanguageCode = htmlspecialchars( $LanguageCode );
 $viewParameters = array( 'offset' => $Offset );
 
+ezpEvent::getInstance()->notify('content/cache/translations', array($ObjectID, $EditVersion, [$LanguageCode]));
```
andrerom: Afaik this is the wrong place to clear cache, and also even name implies it's a cache event, however what user is doing here is not initiate clear cache, but rather viewing a preview.

A bit of a guess, but I would think there needs to be a event (if there isn't already) fired when system stores draft, which is afaik done before redirecting to previewing it. right?

zerustech: @andrerom Thanks for your feedback. I have moved the code to `attribute_edit.php`.

andrerom:
Closing, alternative solution posted in https://github.com/ezsystems/LegacyBridge/pull/153

PR #1376, opened 2018-07-11
EZP-29400 : ezobjectrelationlist - fix fatal error

tharkun:
Fixes a fatal error of the issue #29400 : https://jira.ez.no/browse/EZP-29400


andrerom:
Thanks @tharkun! Released this as part of 2018.06.1

PR #1377, opened 2018-07-16
Fix EZP-29418: Wrong Session Name after EZP-29390

blankse:
Question | Answer
-- | --
JIRA issue | [EZP-29418](https://jira.ez.no/browse/EZP-29418)
Bug | yes
New feature | no
Target version | 2017.12

Regression from https://github.com/ezsystems/ezpublish-legacy/pull/1373
When a sessionName is set, it will generate a new one in the else if section.

### DIFF
--- a/lib/ezsession/classes/ezsession.php
+++ b/lib/ezsession/classes/ezsession.php
@@ -269,9 +269,11 @@ class eZSession
             return false;
 
         $ini = eZINI::instance();
-        if ( $sessionName !== false && $sessionName !== session_name() )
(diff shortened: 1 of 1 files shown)


andrerom:
Fixed CS in c17f617 and released as:
- https://github.com/ezsystems/ezpublish-legacy/releases/tag/v2017.12.3.2
- https://github.com/ezsystems/ezpublish-legacy/releases/tag/v2018.06.1.1


Thanks a lot @blankse ! 👍

blankse:
Thank you @andrerom !

PR #1378, opened 2018-07-19
EZP-29379: Fixed changing of section not committing to the solr index

mateuszbieniek:
| Question           | Answer
| **JIRA issue**     | [EZP-29379](https://jira.ez.no/browse/EZP-29379)
| **Bug/Improvement**| yes
| **Target version** | 5.4

Changing section for content inside Details tab updates the database with change but does not commit to solr index. This is because solr indexer is used inside a transaction that updates section id for the object (and when indexing takes place, sections are not changed) and cache for items in subitem tree are never cleared before indexing (so they end up indexed with outdated data).
This PR fixes the above problems.

### DIFF
--- a/kernel/classes/ezcontentobjecttreenode.php
+++ b/kernel/classes/ezcontentobjecttreenode.php
@@ -2799,25 +2799,13 @@ class eZContentObjectTreeNode extends eZPersistentObject
     static function assignSectionToSubTree( $nodeID, $sectionID, $oldSectionID = false )
     {
         $db = eZDB::instance();
-
         $node = eZContentObjectTreeNode::fetch( $nodeID );
-        $nodePath =  $node->attribute( 'path_string' );
+
+        $objectSimpleIDArray = eZContentObjectTreeNode::getObjectIdsInNodeSubTree($node);
+        if ( !$objectSimpleIDArray )
+            return;
 
         $sectionID =(int) $sectionID;
-
-        $pathString = " path_string like '$nodePath%' AND ";
-
-        // fetch the object id's which needs to be updated
-        $objectIDArray = $db->arrayQuery( "SELECT
-                                                   ezcontentobject.id
-                                            FROM
-                                                   ezcontentobject_tree, ezcontentobject
-                                            WHERE
-                                                  $pathString
-                                                  ezcontentobject_tree.contentobject_id=ezcontentobject.id AND
-                                                  ezcontentobject_tree.main_node_id=ezcontentobject_tree.node_id" );
-        if ( count( $objectIDArray ) == 0 )
-            return;
 
         // Who assigns which section at which node should be logged.
         $section = eZSection::fetch( $sectionID );
@@ -2827,12 +2815,6 @@ class eZContentObjectTreeNode extends eZPersistentObject
                                                       'Content object ID' => $object->attribute( 'id' ),
                                                       'Content object name' => $object->attribute( 'name' ),
                                                       'Comment' => 'Assigned a section to the current node and all child objects: eZContentObjectTreeNode::assignSectionToSubTree()' ) );
-
-        $objectSimpleIDArray = array();
-        foreach ( $objectIDArray as $objectID )
-        {
-            $objectSimpleIDArray[] = $objectID['id'];
-        }
 
         $filterPart = '';
         if ( $oldSectionID !== false )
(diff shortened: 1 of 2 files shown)


kernel/classes/ezcontentobjecttreenode.php
```diff
@@ -2798,41 +2798,22 @@ static function sortArrayBySortFieldAndSortOrder( $sortField, $sortOrder )
     */
     static function assignSectionToSubTree( $nodeID, $sectionID, $oldSectionID = false )
     {
-        $db = eZDB::instance();
+        $objectIDArray = eZContentObjectTreeNode::getObjectIdsInNodeSubTree($nodeID);
```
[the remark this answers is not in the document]
kmadejski: Good catch! @mateuszbieniek maybe you can pass `$node` to the `eZContentObjectTreeNode::getObjectIdsInNodeSubTree` instead of `$nodeID`?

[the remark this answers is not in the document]
mateuszbieniek: @andrerom you are right, the node is fetched twice - I will refactor `getObjectIdsInNodeSubTree` as @kmadejski suggested.
About the risk mentioned earlier: `assignSectionToSubTree` is called inside transaction started here https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/classes/ezsection.php#L219 (it is called inside `eZContentOperationCollection::updateSection` https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/classes/ezsection.php#L243 ), so I think we should be fine.

kernel/classes/ezcontentobjecttreenode.php
```diff
@@ -2842,15 +2823,15 @@ static function assignSectionToSubTree( $nodeID, $sectionID, $oldSectionID = fal
...
+        foreach ( array_chunk( $objectIDArray, 100 ) as $pagedObjectIDs )
         {
             $db->query( "UPDATE ezcontentobject SET section_id='$sectionID' WHERE $filterPart " . $db->generateSQLINStatement( $pagedObjectIDs, 'id', false, true, 'int' ) );
             eZSearch::updateObjectsSection( $pagedObjectIDs, $sectionID );
         }
         $db->commit();
 
         // clear caches for updated objects
-        eZContentObject::clearCache( $objectSimpleIDArray );
+        eZContentObject::clearCache( $objectIDArray );
```
andrerom: I'm still not fully able to see what is changed here besides the variable name (and the two other new issues written above)

Could you expand on what you are fixing in this part of the change _(ezcontentobjecttreenode.php)_?

mateuszbieniek: Sure: in the original code, there is a part where [$objectSimpleIDArray](https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/classes/ezcontentobjecttreenode.php#L2831) is defined only to store id values from another array [$objectIDArray](https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/classes/ezcontentobjecttreenode.php#L2811) which does not contain objects IDs but whole records returned from DB - this is a case of bad variable naming. The code responsible for getting records from DB and creating the array of IDs for it was moved to the `getObjectIdsInNodeSubTree` method, so there is no need for little misleading `$objectSimpleIDArray` variable.

To summarize: the code responsible for `$objectSimpleIDArray` was moved to another method thus this variable (variable name) is no longer needed.

andrerom: So we don't really need to risk changing this part of the code to solve the bug right? ;)

mateuszbieniek:
The code was changed to remove unnecessary refractories and `getObjectIdsInNodeSubTree` works on `$node` inseat `$nodeId` as @kmadejski sugested. Ping @andrerom.

pkamps:
From what I read in the bug report, the problem becomes visible with a permission issue. I tried to reproduce the issue in my installation but I cannot.

Also, I'm sure the section ID is set correctly in solr. There is even a comment about about the possible transaction issue in ezfind `ezsolr.php::updateObjectsSection`. Here is the code to avoid the race condition with the DB transaction:

```
            $object = eZContentObject::fetch( $id );
            // we may be inside a DB transaction running update queries for the
            // section id or the content object may come from the memory cache
            // make sure the section_id is the right one
            $object->setAttribute( 'section_id', $sectionID );
            eZContentOperationCollection::registerSearchObject( $id );
```

With the code changes in this pull request you update the solr index twice:

https://github.com/mateuszbieniek/ezpublish-legacy/blob/dffb0fb022dc39cb0f87a2b4a44132b43cce1dc9/kernel/classes/ezcontentobjecttreenode.php#L2830

https://github.com/mateuszbieniek/ezpublish-legacy/blob/dffb0fb022dc39cb0f87a2b4a44132b43cce1dc9/kernel/classes/ezsection.php#L267

I'd suggest to investigate why ateuszbieniek runs into that permission issue. But I'm pretty sure that the solr section id is correctly set (looking directly at the solr entries).

andrerom:
I think @pkamps is talking about ezfind usage here, and not solr bundle which is what is affected by the bug.

pkamps:
Ah that makes sense. I agree with you to update solr (ezfind, solr bundle) after the DB commit. But I don't think it's good to have the code doing it twice. 
Also, the code probably should be part of the ez content operation "updatesection".

mateuszbieniek:
I don't think that this code can be moved to `updatesection` as it is invoked in `applyTo` method inside db transaction, so changes will be not available for indexing.
As for invoking `updateObjectsSection` twice - it will be hard to remove the first occurrence, as some other code could depend on it. That would result in BC break. Choosing "lesser evil" I decided to not remove this part from the legacy code.

andrerom:
There could be an argument passed to it for flagging if change should be indexed or not, by default it should for BC.

mateuszbieniek:
That's a good idea. I will prepare PR for it later on.

PR #1379, opened 2018-07-20
Throw exception when module or view is not found.

zerustech:
The legacy ezpublish does not return the 404 response code for any
request to a page that does not exist. It simply redirects all invalid
requests to the login page with a resonse code 200.

To fix this issue, method `ezpkernelweb::dispatchLoop()` has been
changed to throw exception when the requested module or view does not
exist.

| Question           | Answer
| **JIRA issue**     | n/a
| **Bug**            | yes
| **New feature**    | no
| **Target version** | `2017.12`
| **BC breaks**      | yes
| **Tests pass**     | yes
| **Doc needed**     | no

**TODO**:
- [ ] Implement feature / fix a bug.
- [ ] Implement tests.
- [ ] Fix new code according to Coding Standards (`$ composer fix-cs`).
- [x] Ask for Code Review.

### DIFF
--- a/kernel/private/classes/ezpkernelweb.php
+++ b/kernel/private/classes/ezpkernelweb.php
@@ -637,6 +637,14 @@ class ezpKernelWeb implements ezpWebBasedKernelHandler
             }
             else
             {
+                $moduleName = $this->uri->element();
+                $viewName = $this->uri->element(1);
+                $module = eZModule::findModule($moduleName);
+                if (null === $module) {
+                    throw new ezpModuleNotFound( $moduleName );
+                } elseif (!isset($module->Functions[$viewName])) {
+                    throw new ezpModuleViewNotFound( $moduleName, $viewName );
+                }
                 $moduleCheck = eZModule::accessAllowed( $this->uri );
             }
 


gggeek:
I am not sure that the described behaviour is what happens by default... or at least I don't remember this happening all the time. Could you give a bit more details about the set up? Is that pure-legacy or eZ5.4 ? Backoffice or frontend ?

zerustech:
Hi @gggeek,
It is an installation of eZ Platform (2.0) + Legacy Bridge (v2.0.6). Legacy mode is enabled for the legacy admin site-access and when you access any invalid URL at the front office, you will be redirected to the legacy login page.
Please also refer to ezsystems/LegacyBridge#155 for details.

zerustech:
@gggeek The cause of this problem is that an invalid URL can NOT be parsed by any routers, so it falls back to the legacy controller, which passes the URL to the `ezpkernelweb` class. Currently, this class does not check if the URL maps to any view of any module, it simply checks if current user has the permission to access this URL, and if not (could be the case that the URL does not map to any valid module or view at all), it just redirects to the login page and does not return the 404 error code.

emodric:
@zerustech Where did you test this behaviour? On frontend (with legacy mode enabled or disabled) or backend (again with legacy mode enabled or disabled)? Because I can't seem to reproduce it.

I have two siteaccesses:

1) Frontend with legacy mode disabled - no issue. `/content/abc` shows a Symfony Not Found page
2) Legacy admin backend with legacy mode enabled - no issue. To anonymous users, login page is shown without redirect, but this is expected (because of `RequireUserLogin=true`), and to logged in users, legacy error page is shown.

zerustech:
@emodric I also have two siteaccesses:
1. Frontend (legacy mode disabled) - `/content/abc` shows "legacy login page"
1. Legacy admin backend (legacy mode enabled) - works as expected (same as your testing result).

zerustech:
@emodric By the way, the version of `ezpublish-legacy` is `v2018.06.1.1`.

emodric:
Can you share the details of your frontend siteaccess configuration (both in legacy and new stack), or a consistent way to reproduce the issue (maybe from a clean install, since I can't reproduce it unfortunatelly?

andrerom:
> @andrerom Any input?

hmm..

in this specific case maybe we should inject the config for this one, and maybe there are other cases like it.

emodric:
Maybe. The problem is that it should be injected only for some siteaccesses (frontend ones), unlike view cache.

andrerom:
True, but what does it very on? is there some setting in Platform we can safely derived this from?

emodric:
I don't think so. Admin siteaccesses (legacy, ngadmin) need it turned on, frontend ones don't.

andrerom:
hmm, in v2 we _could_ detect admin group, right? But then we should document this as a convention.

PR #1380, opened 2018-07-30
[WIP] EZP-28681: Legacy transaction handling issues

glye:
> Fixes https://jira.ez.no/browse/EZP-28681
> Work in progress

- [x] 1. Status change during autosave, see https://github.com/ezsystems/ezautosave/pull/27
- [ ] 2. Attributes deleted during draft discard https://github.com/ezsystems/ezpublish-legacy/pull/1380/commits/ed5acf098a5de92249633347fde3543634261226
- [ ] 3. Status change during draft discard https://github.com/ezsystems/ezpublish-legacy/pull/1380/commits/68b632e6795cd362d6434f47a37c0b449d34b634
- [x] 4. Current version changed during publish https://github.com/ezsystems/ezpublish-legacy/pull/1380/commits/bd7bcb8866819c8b235bf45babb10266a3a3c2ab
- [ ] 5. Files deleted during DFS cache purge https://github.com/ezsystems/ezpublish-legacy/pull/1380/commits/ab9bdae0fcd782a52c99626cf9e272220458f4f2 Problematic, see https://github.com/ezsystems/ezpublish-legacy/pull/1380#pullrequestreview-141520631

### DIFF
--- a/kernel/private/classes/clusterfilehandlers/dfsbackends/mysqli.php
+++ b/kernel/private/classes/clusterfilehandlers/dfsbackends/mysqli.php
@@ -334,7 +334,6 @@ class eZDFSFileHandlerMySQLiBackend implements eZClusterEventNotifier
      * @see _purge
      *
      * @return bool|int false if it fails, number of affected rows otherwise
-     * @todo This method should also remove the files from disk
      */
     public function _purgeByLike( $like, $onlyExpired = false, $limit = 50, $expiry = false, $fname = false )
     {
@@ -392,11 +391,15 @@ class eZDFSFileHandlerMySQLiBackend implements eZClusterEventNotifier
             return $this->_fail( "Purging file metadata by like statement $like failed" );
         }
         $deletedDBFiles = mysqli_affected_rows( $this->db );
-        $this->dfsbackend->delete( $files );
 
-        $this->_commit( $fname );
+        if ( $this->_commit( $fname ) === true )
+        {
+            $this->dfsbackend->delete( $files );
 
-        return $deletedDBFiles;
+            return $deletedDBFiles;
+        }
+
+        return false;
     }
 
     /**
@@ -1351,6 +1354,10 @@ class eZDFSFileHandlerMySQLiBackend implements eZClusterEventNotifier
     /**
      * Stops a current transaction and commits the changes by executing a COMMIT call.
      * If the current transaction is a sub-transaction nothing is executed.
+     *
+     * Returns true if the commit was executed successfully, false if it failed, null if it wasn't executed.
+     *
+     * @return bool|null
      */
     protected function _commit( $fname = false )
     {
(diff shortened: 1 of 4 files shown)


gggeek:
See comments in ezsystems/ezautosave#27 ...

kernel/private/classes/clusterfilehandlers/dfsbackends/mysqli.php
```diff
@@ -392,11 +391,15 @@ public function _purgeByLike( $like, $onlyExpired = false, $limit = 50, $expiry
             return $this->_fail( "Purging file metadata by like statement $like failed" );
         }
         $deletedDBFiles = mysqli_affected_rows( $this->db );
-        $this->dfsbackend->delete( $files );
 
-        $this->_commit( $fname );
+        if ( $this->_commit( $fname ) === true )
```
glye: This is problematic, I know. Aside from the issue of adding a return value to `_commit`, what happens if the commits are recursive and `null` is returned? The files won't be deleted, but if the outer commit eventually goes through, we would want them to be. OTOH if we did delete them in the null case, and the outer commit was rolled back, we'd have the opposite problem.

glye:
Update for 4. Current version changed during publish
Lock the affected rows in share mode. Means we must query ezcontentobject even when `$versionNum` is specified.

glye:
Update for 4. Current version changed during publish
Lock the affected rows in share mode. Means we must query ezcontentobject even when $versionNum is specified. This Mysql specific and must be moved to the eZDB abstraction layer. Just testing it here now.

gggeek:
difference between 'for update' and 'lock in shared mode': https://stackoverflow.com/questions/32827650/mysql-innodb-difference-between-for-update-and-lock-in-share-mode/38729206

glye:
@gggeek Yes...? Nice to know, but I'm afaik not creating any rows in the commit for point 4, so don't see a problem with counters. I am writing to ezcontentobject_version, but didn't see any reason to lock it for reading. If someone else is trying to set the status of the same version at the same time, it shouldn't matter if they read the old or the new row data - the action doesn't depend on the status. But chances are I've missed something :)

gggeek:
The thing is, if you do 'select for update' you will not prevent others to do plain selects. But others transactions having the same select for update' will queue, which is what we want.
Reading that SO page makes me thing that using 'lock in share mode' will not prevent another update transaction to start its operation

gggeek:
Correct. Generally speaking, you want to lock as few tables as possible, to avoid the risk of ending up with the 'dining philosophers' situation, where threads deadlock each other.
The 'select for update' that you should be:
- on the main table holding the data you want to update (in this case most likely version and/or object)
- used to ascertain that the state of the data is good top start the transaction (ie. no checking of version status in a previous select, as that would not be atomic)

ps: seems that mysql at least supports select-for-update on a join... https://stackoverflow.com/questions/38544770/mysql-can-i-use-one-select-for-update-to-protect-multiple-tables-lock

glye:
I locked "for update" in the second query now since I found a hard reason to do so: A read lock doesn't stop other sessions from also getting a read lock. That other lock would block my update.

Now to find a way to do this properly within the eZDB framework... :hurtrealbad:

glye:
Closing. I'll deal with these in separate PRs instead:
1. https://github.com/ezsystems/ezautosave/pull/27
2. https://github.com/ezsystems/ezpublish-legacy/pull/1397
3. https://github.com/ezsystems/ezpublish-legacy/pull/1398
4. https://github.com/ezsystems/ezpublish-legacy/pull/1399
5. https://github.com/ezsystems/ezpublish-legacy/pull/1400

PR #1384, opened 2018-08-17
[WIP] EZP-29279: Values of Date Field Type should be handled in UTC only

jacek-foremski:
JIRA issue: [EZP-29279](https://jira.ez.no/browse/EZP-29279)

Currently, in Legacy the timestamp stored in the database is always midnight in the server's timezone, while in Platform it is always in UTC. This PR changes how values of Date Field Type are handled to make the behavior consistent with Platform. It also makes sure that the displayed values are also in UTC and unaffected by server's timezone. 

**TODO:**
- [ ] Migration script that changes values saved in the database to UTC.

### DIFF
--- a/lib/ezlocale/classes/ezdate.php
+++ b/lib/ezlocale/classes/ezdate.php
@@ -62,6 +62,8 @@ class eZDate
      */
     public function __construct( $date = false )
     {
+        $tz = date_default_timezone_get();
+        date_default_timezone_set( 'UTC' );
         if ( $date === false )
         {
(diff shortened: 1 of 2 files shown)


mateuszbieniek:
@andrerom I agree - eZDate is used in other places, so I prepared new PR with a different approach:
https://github.com/ezsystems/ezpublish-legacy/pull/1401

andrerom:
Closing in favour of #1401 as this PR is too intrusive on lib classes used several places in the system.

PR #1386, opened 2018-08-28
EZP-29573: Legacy installer should allow utf8mb4 charset

glye:
> Fixes https://jira.ez.no/browse/EZP-29573
> Ready for review

When creating a database in `utf8mb4`, the legacy installer complains that only `utf-8` is supported. Since `utf8mb4` is recommended now we should support both.

`utf8mb4` is not available in every database eZ Publish 5.x supports, but then people won't be able to create such a DB, so it won't be a problem.

**NB:** Since we have not backported the utf8mb4 schema changes, in order for this to work one has to first create the database, then apply the changes in the [advisory](http://share.ez.no/community-project/security-advisories/ezsa-2018-003-4-byte-utf-8-in-mysql-mariadb), and then run the installer.

### DIFF
--- a/kernel/setup/steps/ezstep_installer.php
+++ b/kernel/setup/steps/ezstep_installer.php
@@ -489,12 +489,11 @@ class eZStepInstaller
             $checkedCharset = $db->checkCharset( $charsetsList, $currentCharset );
             if ( $checkedCharset === false )
             {
-                // If the current charset is utf-8 we use that instead
-                // since it can represent any character possible in the chosen languages
-                if ( $currentCharset == 'utf-8' )
+                // If the current charset is utf-8 or utf8mb4 we use that instead
+                // since they can represent any character possible in the chosen languages
+                if ( in_array( $currentCharset, array( 'utf-8', 'utf8mb4' ) ) )
                 {
-                    $charset = 'utf-8';
-                    $result['site_charset'] = $charset;
+                    $result['site_charset'] = $currentCharset;
                 }
                 else
                 {
--- a/kernel/setup/steps/ezstep_installer.php
+++ b/kernel/setup/steps/ezstep_installer.php
@@ -493,7 +493,7 @@ class eZStepInstaller
                 // since they can represent any character possible in the chosen languages
                 if ( in_array( $currentCharset, array( 'utf-8', 'utf8mb4' ) ) )
                 {
-                    $result['site_charset'] = $currentCharset;
+                    $result['site_charset'] = 'utf-8';
                 }
                 else
                 {


kernel/setup/steps/ezstep_installer.php
```diff
@@ -489,12 +489,11 @@ function checkDatabaseRequirements( $dbCharset = false, $overrideDBParameters =
             $checkedCharset = $db->checkCharset( $charsetsList, $currentCharset );
             if ( $checkedCharset === false )
             {
-                // If the current charset is utf-8 we use that instead
-                // since it can represent any character possible in the chosen languages
-                if ( $currentCharset == 'utf-8' )
+                // If the current charset is utf-8 or utf8mb4 we use that instead
+                // since they can represent any character possible in the chosen languages
```
glye: Disclaimer: Not 100% true, certain exotic natural language characters are not possible in utf-8, but in most cases the reason utf8mb4 is needed is security and emoji.

andrerom:
@glye did you figure out anything on this?

glye:
@andrerom No, delayed by other things. Need to get back to it.

andrerom:
@vidarl was this one of the issue you where hitting when trying to use setup wizard?

vidarl:
> @vidarl was this one of the issue you where hitting when trying to use setup wizard?

@andrerom : no. My problem was that the setup wizard would refuse to use a database which has the 'utf8mb4' character set... So after choosing the database to use (in the drop-down and click 'next'), it would say 'Hey.... this db is not in "utf8"'

andrerom:
@vidarl that sounds exactly like the description in PR here :)

vidarl:
hehe.. that is right.. sorry. I was a bit quick and only read micszo's [comment](https://github.com/ezsystems/ezpublish-legacy/pull/1386#pullrequestreview-152502311)....

micszo:
btw the latest status is in Jira :) 
https://jira.ez.no/browse/EZP-29573?focusedCommentId=233444&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-233444

kernel/setup/steps/ezstep_installer.php
```diff
@@ -489,12 +489,11 @@ function checkDatabaseRequirements( $dbCharset = false, $overrideDBParameters =
...
+                if ( in_array( $currentCharset, array( 'utf-8', 'utf8mb4' ) ) )
                 {
-                    $charset = 'utf-8';
-                    $result['site_charset'] = $charset;
+                    $result['site_charset'] = $currentCharset;
```
andrerom:
```suggestion
                    $result['site_charset'] = 'utf-8';
```

As this option is in the end passed on to `i18n.ini` it seems like it might need to be set to utf-8 even if we got utf8mb4 as input. Or do we need to adapt the whole setup code base to support this value?

glye: @andrerom Right, we only need `utf8mb4` when talking to mysql. We can test it.

glye:
@micszo Hi! Could you test this again with the last change from André, please?

micszo:
@glye tried again on v5.4.12.1 with latest diff from this PR, followed steps as in my review from Sep 5. Getting the same results, 504 on publish etc.

Are there other places in the system apart from i18n.ini (i18n.ini.append.php) and ezpublish.yml where utf-8 should be changed to utf8mb4?
Maybe we need a more detailed instruction for this.

micszo:
OK, breakthrough. 
After completing all the steps I dropped the db, created it again, forced the setup wizard again, then applied the script from the advisory. Voilà. All issues gone.
Seems like the is some issue with sequence here.

glye:
@micszo Awesomtastic! So what was the difference in sequence from the previous test?

micszo:
What I did.

dropped db, created again
rm -rf * in install dir
cloned meta
composer install tag
setup wizard
diff from PR (to unblock wizard)
script from advisory
updated charset (site.ini, i18n.ini, ezpublish.yml)
reloaded cache (admin and cli)

dropped db, created again
setup wizard
script from advisory
reloaded cache

Dunno why it didn't work without repeating those 4 steps.

micszo:
I'm still uncertain about this because second wizard wiped some of those charset changes. Checking.

glye:
I looked through it again, and it seems correct that there should be no change in i18n.ini. Is there any reference to i18n.ini in the PR or in the jira issue, that I missed?

micszo:
Nope, can't see it either. Must have looked for all places where charset was present. In that case, if it's not required there, we might be good to go. 😅

glye:
Yep, as I recall the i18n.ini change was a mistake, as that setting is about the output charset. The site.ini setting is for the database charset, which is what we want here. `utf8mb4` is a mysql-only charset, for output we use utf-8.

PR #1388, opened 2018-09-07
EZP-29598: Prevent race condition when generating TS cache

omh:
Ticket: https://jira.ez.no/browse/EZP-29598

### DIFF
--- a/lib/ezi18n/classes/eztstranslator.php
+++ b/lib/ezi18n/classes/eztstranslator.php
@@ -218,6 +218,10 @@ class eZTSTranslator extends eZTranslatorHandler
                                 $this->Messages[$key] = $msg;
                             }
                         }
+                        else
+                        {
+                            $this->BuildCache = true;
+                        }
                         eZDebug::accumulatorStop( 'tstranslator_context_load' );
                     }
                 }


ernestob:
Hi guys,

I have upgraded a multi-language site, this change is breaking the whole site.

After I applied this change, there was a serious performance problem. The website pages were taking about 10-20 minutes to render --if the pages were rendered at all.

Removing this change returned the site's regular speed.

glye:
(We will follow up this in the CS issue reported by Peter.)

omh:
Hi all,

Just to chime in that we've been running this exact patch (no upgrade, just the patch) since the pull request was made with no known problems. This is on a fairly heavy traffic site. There's enough traffic that we can't deploy during normal office hours without taking the site down (this was the case even before the patch).

@ernestob Did you roll back this exact change (4 lines) or did you roll back to a previous eZ version?

omh:
Yes, we do run *this* patch as we applied the patch we got via support portal, but I logged in to the server and double checked anyways 👇

screenshot 2018-12-07 at 08 56 48

I think perhaps @andrerom is onto something, the logic in `canRestoreCache` does seem a bit off compared to how we assume it works in the patch. If that theory is correct cache generation might indeed be more inefficient now and explain @ernestob issues. Once cached it should be behave as before.

andrerom:
@omh Could be you have faster disk so you don't notice so much the extra writes being performed on the installs you have? Or maybe Cluster / not-cluser affects this?

Either way, are you ok from your side that we aim to revert back to your original patch and do testing on that internally?

omh:
@andrerom Hm, think it's a fairly old server so wouldn't expect it to be that fast. It is dedicated hardware though, not cloud stuff. We also run in a cluster via NFS.

Seems reasonable to revert back and test some more and see if both issues can be solved at the same time :).

andrerom:
@omh We never reproduced the race condition, so we'll need to rely on your input on that.

konradoboza:
According to multiple tests, I wasn't able to observe translation cache files regeneration on each request. I tested pages containing strings translated via `i18n` operator. After initial cache warm-up, I cleared view cache and tried to refresh the site and also navigate to other pages which contain translated strings. Translation cache files weren't modified. I tested on eZ Publish 5.4.13 with demo content and additional translated strings. @ernestob could you please provide us with some more details regarding your setup? Maybe I am missing some important setting or step to reproduce the issue.

PR #1389, opened 2018-09-07
Atomic cache writes to prevent corrupt caches.

omh:
Some caches doesn't use atomic writes when writing to disk. This can lead to corrupt cache files if multiple processes tries to write to the same file at the same time.


andrerom:
@omh Have you seen similar issues also with template compile cache?

omh:
@andrerom No concrete issues that I have managed to track down no. Stumbled across these while reproducing/fixing #1388.

PR #1395, opened 2018-10-22
[PHP7] Add ctor compat on eZContentObjectTreeNode

andrerom:


emodric:
@andrerom Is there a specific reason we're adding this one back?

gggeek:
To insure compat for custom code of customers?

emodric:
@gggeek I know that :) I meant, is there some open source code out there (official or not) which breaks without it?

andrerom:
It's for cases like this: https://github.com/ezsystems/ezfind/commit/1d2076763e5ab58fe2ba5a38b0308505a0ff88fc

So it's a bit overtime, but had a person running into it again recently since they had forgotten to upgrade ezfind.

PR #1396, opened 2018-10-24
fixed fatal error in eZUserType::objectAttributeContent

pbek:
Fixed a fatal error in eZUserType::objectAttributeContent because user object could not be loaded

### DIFF
--- a/kernel/classes/datatypes/ezuser/ezusertype.php
+++ b/kernel/classes/datatypes/ezuser/ezusertype.php
@@ -364,6 +364,11 @@ class eZUserType extends eZDataType
         $user = eZUser::fetch( $userID );
         eZDebugSetting::writeDebug( 'kernel-user', $user, 'user' );
 
+        // return if user object wasn't found to avoid a fatal error in eZUserType::updateUserDraft
+        if ( !( $user instanceof eZUser ) ) {
+            return null;
+        }
+
         // Looking for a "draft" and loading it's content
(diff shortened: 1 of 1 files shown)


pbek:
> Opening curly brace needs to go on own line

done

glye:
Unrelated to the fix, but isn't it odd that `hasObjectAttributeContent()` will return false if the user exists but is disabled?
https://github.com/ezsystems/ezpublish-legacy/blob/c306576fbc39644bbda26d9fd0f29eba82fc5e77/kernel/classes/datatypes/ezuser/ezusertype.php#L347

andrerom:
@pbek Seems there is a conflict here now after https://github.com/ezsystems/ezpublish-legacy/pull/1417, is this still needed?

pbek:
I can't see this anywhere fixed in #1417, so I guess this is still needed.

andrerom:
ok, would you be able to rebase @pbek ?

andrerom:
@pbek Is the update her ok with you? And is there a JIRA issue for this?

pbek:
Looks ok to me.

> And is there a JIRA issue for this?

Not that I know of.

andrerom:
@pbek ok, can you add one with info on when you encountered this so we have it for anyone looking up similar case?

andrerom:
Dropping from 2019.03 milstone until it is clear what this actually fixes.

PR #1400, opened 2018-11-12
EZP-29800: Legacy transaction handling: DFS cache purge

glye:
> Fixes https://jira.ez.no/browse/EZP-29800
> Sub-task of https://jira.ez.no/browse/EZP-28681
> See history in obsolete shared PR https://github.com/ezsystems/ezpublish-legacy/pull/1380

We need to know if the commit succeeded, before we can delete files.  This adds a return value to `_commit()` for that purpose.

This is problematic. What happens if the commits are recursive and `null` is returned? The files won't be deleted, but if the outer commit eventually goes through, we would want them to be. OTOH if we did delete them in the null case, and the outer commit was rolled back, we'd have the opposite problem.

Then again, it looks to me like the way purge is called, there won't be any recursive transactions here, so it's no issue in practice. The important thing is knowing if the commit failed, and not delete any files in that case. And that is covered now.

### DIFF
--- a/kernel/private/classes/clusterfilehandlers/dfsbackends/mysqli.php
+++ b/kernel/private/classes/clusterfilehandlers/dfsbackends/mysqli.php
@@ -334,7 +334,6 @@ class eZDFSFileHandlerMySQLiBackend implements eZClusterEventNotifier
      * @see _purge
      *
      * @return bool|int false if it fails, number of affected rows otherwise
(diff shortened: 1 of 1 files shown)


glye:
Testing out the null case.

glye:
Closing, merged in https://github.com/ezsystems/ezpublish-legacy/pull/1415

PR #1401, opened 2018-11-29
EZP-29279: Values of Date Field Type should be handled in UTC only

mateuszbieniek:
JIRA issue: [EZP-29279](https://jira.ez.no/browse/EZP-29279)
Replaces #1384

Currently, in Legacy, the timestamp stored in the database is always midnight in the server's timezone, while in Platform it is always in UTC.
This PR modifies `eZDateType` class so timestamps are converted to UTC before being stored and are converted back to Local Timezone timestamp on retrieval.

### DIFF
--- a/lib/ezutils/classes/eztimestamp.php
+++ b/lib/ezutils/classes/eztimestamp.php
@@ -1,0 +1,43 @@
+<?php
+/**
+ * File containing the eZTimestamp class.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ * @package lib
+ */
+
+class eZTimestamp
+{
+    /*!
+     \return a timestamp in UTC
+    */
+    public static function getUtcTimestampFromLocalTimestamp($localTimestamp) {
+        $utcTimezone = new \DateTimeZone('UTC');
+        $localTimezone = new \DateTimeZone(date_default_timezone_get());
+
+        $localDate = new \DateTime(null, $localTimezone);
+        $localDate->setTimestamp($localTimestamp);
+
+        $utcDate = new \DateTime($localDate->format('Y-m-d'), $utcTimezone);
+        return $utcDate->getTimestamp();
+    }
+
+    /*!
+     \return a timestamp in timezone defined in php.ini
+    */
+    public static function getLocalTimestampFromUtcTimestamp($utcTimestamp) {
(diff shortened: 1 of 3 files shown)


mateuszbieniek:
Thanks @glye! DataTime has different JIRA issue and I will be creating similar PR for it after this one is done ¯\_(ツ)_/¯

lib/ezutils/classes/eztimestamp.php
```diff
@@ -0,0 +1,43 @@
...
+ * @version //autogentag//
+ * @package lib
+ */
+
+class eZTimestamp
+{
+    /*!
+     \return a timestamp in UTC
+    */
+    public static function getUtcTimestampFromLocalTimestamp($localTimestamp) {
```
glye: Strictly speaking, this file needs space between parentheses everywhere for legacy CS: `foo( $bar )`

mateuszbieniek:
```
        ____________________
  ___  |        FRESH       |
 /_| | |       SPACES       |
|    |_|__________________  |
"-O----O-O' `         `O`O'-'
```

lib/ezutils/classes/eztimestamp.php
```diff
@@ -0,0 +1,43 @@
...
+        $utcTimezone = new \DateTimeZone( 'UTC' );
+        $localTimezone = new \DateTimeZone( date_default_timezone_get() );
+
+        $utcDate = new \DateTime( null, $utcTimezone );
+        $utcDate->setTimestamp( $utcTimestamp );
+
+        $localDate = new \DateTime( $utcDate->format( 'Y-m-d H:i:s' ), $localTimezone );
+        $localTimestamp = $localDate->getTimestamp();
+
+        return $localTimestamp;
```
kmadejski: Maybe you should do it like:
```
$localDate = new \DateTime( $utcDate->format( 'Y-m-d H:i:s' ), $localTimezone );

return $localDate->getTimestamp();
```
or even:
```
return new \DateTime( $utcDate->format( 'Y-m-d H:i:s' ), $localTimezone )->getTimestamp();
```
but I prefer the first way, it's more readable. 

The same applies for https://github.com/ezsystems/ezpublish-legacy/pull/1401/files/6b98fbe6e7b1c9e93bbf6b42a1d00205f42705bd#diff-9ae7112c632bad6c06997d57aefd6f82R24.

mateuszbieniek: You are right! I prefer the first approach as well. Fixed.

PR #1402, opened 2018-11-29
 [ezpublish legacy] include assigned nodes into the removethis process in ezcontentobject php class

FatalError404:
* requirement
** node id is used as parameter
** used node id is the main one
* context : known context is the unpublish feature


FatalError404:
it doesn't work to change base 
need to make a new PR using the 2017.12 base

PR #1409, opened 2018-12-19
EZP-29901: Fix Translation cache reloading all the time causing performance issues

andrerom:
https://jira.ez.no/browse/EZP-29901
Followup on #1388

### DIFF
--- a/lib/ezi18n/classes/eztranslationcache.php
+++ b/lib/ezi18n/classes/eztranslationcache.php
@@ -117,9 +117,20 @@ class eZTranslationCache
 
     /*!
      \static
+     \return true if the cache with the key \a $key is in memory already.
+    */
+    public static function hasRestoredCache( $key )
+    {
+        return isset( $GLOBALS['eZTranslationCacheTable'][$key] );
+    }
+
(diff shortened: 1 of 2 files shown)


andrerom:
@ernestob / @peterkeung something that would work for your setup too?

@omh I'm assuming your original intention is still intact here?

peterkeung:
@ernestob can test on his project, but is away until Jan 7

omh:
Will try and test it and reproduce race condition. Might not get to it until after new years though.

andrerom:
FYI: There are two reports (internal and external)that this patch improves the situation, but there are still writes where there were none before #1388.

In our internal findings we observed that there language fallback _might_ be involved here. Code seems to be trying to look for cache of one language which [never exits](https://github.com/ezsystems/ezpublish-legacy/blob/v2018.09.1.3/lib/ezutils/classes/ezphpcreator.php#L856), but in the end always ending up using eng-us (for writes).

andrerom:
#1388 reverted in 7554cf0, closing thus closing this issue but reopening the originally issue instead.

PR #1410, opened 2018-12-28
FIX Bug : call count function on non countable variable

rhidja:


andrerom:
Todo: Rebase on 2017.12 branch.

kernel/classes/ezpersistentobject.php
```diff
@@ -920,7 +920,7 @@ public static function fetchObjectList( $def,
             {
                 $db_params["limit"] = $limit["limit"];
             }
-            else
+            elseif( isset( $limit["length"] ) )
```
andrerom: logic of this change does not make sense, the if is already using this `if ( isset( $limit['limit'] ) )`

andrerom:
Fixed in https://github.com/ezsystems/ezpublish-legacy/commit/215ce2c47a0cd02e91c251e131e3d1c84f04c9a2

PR #1411, opened 2019-01-02
EZP-29957: Two XML blocks in the same Class in a different Categories will break second Online Editor

mateuszbieniek:
JIRA issue: https://jira.ez.no/browse/EZP-29957

This PR backports a fix applied in the 4.x version of TinyMCE:
https://github.com/tinymce/tinymce/blob/4.2.x/js/tinymce/classes/dom/Selection.js#L605


mateuszbieniek:
@andrerom done

PR #1412, opened 2019-01-03
EZP-29723: Content object attributes not set properly

pbek:
This pull request solves <https://jira.ez.no/browse/EZP-29723> while still fixing <https://jira.ez.no/browse/EZP-23088>.

### DIFF
--- a/kernel/classes/datatypes/ezimage/ezimagetype.php
+++ b/kernel/classes/datatypes/ezimage/ezimagetype.php
@@ -190,6 +190,7 @@ class eZImageType extends eZDataType
             $imageHandler->setAttribute( 'alternative_text', false );
             $imageHandler->removeAliases();
             $imageHandler->store( $contentObjectAttribute );
+            $contentObjectAttribute->setContent( null );
         }
     }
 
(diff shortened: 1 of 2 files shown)


kernel/classes/datatypes/ezimage/ezimagetype.php
```diff
@@ -190,6 +190,7 @@ function deleteStoredObjectAttribute( $contentObjectAttribute, $version = null )
             $imageHandler->setAttribute( 'alternative_text', false );
             $imageHandler->removeAliases();
             $imageHandler->store( $contentObjectAttribute );
+            $contentObjectAttribute->setContent( null );
```
glye: Nice, thanks! Looks like the proper way to fix this. I'm only wondering why you're nulling the content _after_ the imagehandler store call? Wouldn't before make more sense?

glye:
Would also prefer commit title like: `EZP-29723: Content object attributes not set properly`
But we can do that on merge.

kernel/classes/datatypes/ezimage/ezimagetype.php
pbek: Everything was stored properly as far as I could see. The issue was the content object attribute object in memory because of the "objects are always references"-thing in PHP (at least that what my "hunch" said)

giRobert:
Hey there, when will this be merged into master?

pbek:
@andrerom are you asking me to do a pull request to https://github.com/ezsystems/ezpublish-legacy/tree/2017.12? I would manage.

pbek:
where you are: #1416

PR #1413, opened 2019-01-03
Add the Copy button to the full view in the Administration Interface

thiagocamposviana:
https://jira.ez.no/browse/EZP-29983

In the eZ back-end you have Edit / Move / Remove on the page you're viewing

But to copy, you have to initiate it from the "Sub items" window. No more! We will add the Copy button to the full view as well!

### DIFF
--- a/design/admin/stylesheets/pagelayout.css
+++ b/design/admin/stylesheets/pagelayout.css
@@ -472,10 +472,11 @@ ul.leftmenu-items li.current a
 #widthcontrol-handler
 {
     position: absolute;
-    right: 20px;
-    top: 0;
+    right: -30px;
+    top:-20px;
     bottom: 0; /* strech div to full width of relative box */
     width: 10px;
+    height: 5000px;
     cursor: col-resize;
     z-index:11;
 /*    background: #f8f8f8;*/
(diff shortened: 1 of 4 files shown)


brookinsconsulting:
@thiagocamposviana Thank you! We have implemented this feature before increasing user friendly ease of use. This is a improvement that we think should be default. +1

Cheers,
Brookins Consulting

design/admin/stylesheets/pagelayout.css
```diff
@@ -472,10 +472,11 @@ ul.leftmenu-items li.current a
 #widthcontrol-handler
 {
     position: absolute;
-    right: 20px;
-    top: 0;
+    right: -30px;
+    top:-20px;
     bottom: 0; /* strech div to full width of relative box */
     width: 10px;
+    height: 5000px;
```
andrerom: Are these changes really relevant here?

thiagocamposviana: @andrerom good catch, probably came from somewhere else, going to update this PR during the weeekend and remove those changes.

andrerom: 😃

PR #1416, opened 2019-01-10
EZP-29723: Content object attributes not set properly

pbek:
This pull request solves <https://jira.ez.no/browse/EZP-29723> while still fixing <https://jira.ez.no/browse/EZP-23088>.

Ported to 2017.12 as asked for in #1412.

### DIFF
--- a/kernel/classes/datatypes/ezimage/ezimagetype.php
+++ b/kernel/classes/datatypes/ezimage/ezimagetype.php
@@ -190,6 +190,7 @@ class eZImageType extends eZDataType
             $imageHandler->setAttribute( 'alternative_text', false );
             $imageHandler->removeAliases();
             $imageHandler->store( $contentObjectAttribute );
+            $contentObjectAttribute->setContent( null );
         }
     }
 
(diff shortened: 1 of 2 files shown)


pbek:
ping @andrerom

glye:
I'll merge it, just double checking first that the failing test isn't relevant.

pbek:
Great, thank you!

pbek:
@glye will the changes also be present in the `master` branch some day?

glye:
@pbek They are, since the merge :) https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/classes/datatypes/ezimage/ezimagetype.php#L193
We merge from oldest supported branch (2017.12) upwards to master.

pbek:
Ah, great! :smile:

thiagocamposviana:
**Reporting a potential issue:** 

This patch has caused some regressions. We have a custom datatype whose data is empty after you save a draft. We are not sure of the exact cause, but restoring the code kernel/content/attribute_edit.php:

```
    // While being fetched, attributes might have been modified.
    // The list needs to be refreshed so it is accurately displayed.
    $contentObjectAttributes = $version->contentObjectAttributes( $EditLanguage );
```

We also have a custom edit handler that can no longer save content.

PR #1417, opened 2019-01-17
EZP-29096: Do not store empty draft values in DB

pkamps:
The issue and steps to reproduce it are described here:
https://jira.ez.no/browse/EZP-29096

Or use those steps:

* Create a new user account
* Do not fill any values in the form
* Close the browser tab
* Try to create a new user account
* You will get a transaction error

This patch will prevent the user creation to store any draft data into the DB unless the editor specified a user login.

### DIFF
--- a/kernel/classes/datatypes/ezuser/ezuser.php
+++ b/kernel/classes/datatypes/ezuser/ezuser.php
@@ -228,7 +228,7 @@ class eZUser extends eZPersistentObject
     /**
      * Only stores the entry if it has a Login value
      *
-     * @param null $fieldFilters
+     * @param mixed|null $fieldFilters
      */
     public function store( $fieldFilters = null )
     {
--- a/kernel/classes/datatypes/ezuser/ezuser.php
+++ b/kernel/classes/datatypes/ezuser/ezuser.php
@@ -233,7 +233,11 @@ class eZUser extends eZPersistentObject
         unset( $GLOBALS['eZUserObject_' . $userID] );
         $GLOBALS['eZUserObject_' . $userID] = $this;
         self::purgeUserCacheByUserId( $userID );
-        parent::store( $fieldFilters );
+
+        if( $this->Login )
+        {
+            parent::store( $fieldFilters );
+        }
     }
 
     function originalPassword()
@@ -307,6 +311,11 @@ class eZUser extends eZPersistentObject
         }
     }
 
+    /**
+     * @param integer $id
+     * @param bool $asObject
+     * @return eZUser|null
+     */
     static function fetch( $id, $asObject = true )
     {
         if ( !$id )
--- a/kernel/classes/datatypes/ezuser/ezuser.php
+++ b/kernel/classes/datatypes/ezuser/ezuser.php
@@ -225,7 +225,12 @@ class eZUser extends eZPersistentObject
         return new eZUser( $row );
     }
 
-    function store( $fieldFilters = null )
+    /**
+     * Only stores the entry if it has a Login value
+     *
+     * @param null $fieldFilters
+     */
+    public function store( $fieldFilters = null )
     {
         $this->Email = trim( $this->Email );
         $userID = $this->attribute( 'contentobject_id' );
--- a/kernel/classes/datatypes/ezuser/ezusertype.php
+++ b/kernel/classes/datatypes/ezuser/ezusertype.php
@@ -258,10 +258,10 @@ class eZUserType extends eZDataType
             }
 
             // saving information in the object attribute data_text field to simulate a draft
-            // only if the object version is a draft (status == 0)
+            // only if the object version is a draft
             if (
                 $user->Login &&
-                $contentObjectAttribute->attribute( 'object_version' )->attribute( 'status' ) == 0
+                $contentObjectAttribute->attribute( 'object_version' )->attribute( 'status' ) == eZContentObjectVersion::STATUS_DRAFT
             )
             {
                 $contentObjectAttribute->setAttribute( 'data_text', $this->serializeDraft( $user ) );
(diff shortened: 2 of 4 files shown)


andrerom:
Nice.

We should aim to place this one 2017.12 and up. As for Travis, unsure why but tests seems to fail with:
```
.PHP Fatal error:  Call to a member function store() on null in /home/travis/build/ezsystems/ezpublish-legacy/kernel/classes/datatypes/ezuser/ezusertype.php on line 282
```

kernel/classes/datatypes/ezuser/ezusertype.php
```diff
@@ -258,7 +258,10 @@ function storeObjectAttribute( $contentObjectAttribute )
             }
 
             // saving information in the object attribute data_text field to simulate a draft
-            $contentObjectAttribute->setAttribute( 'data_text', $this->serializeDraft( $user ) );
+            if( $user->Login )
```
glye:
```suggestion
            if ( $user->Login )
```

glye:
onPublish() fails because there is no user object. Seems the code expects the draft to have been stored, so bigger changes are needed.
https://github.com/ezsystems/ezpublish-legacy/pull/1417/files#diff-7c45a8ca23a28df173a64ef5548180c0R282

peterkeung:
Are tests passing with the latest commit?

glye:
@peterkeung No, sorry. Don't you have access to see the results? Travis says `eZUserTypeRegression::testUpdatePasswordUpdatesSerializedData
The password hash stored in the user attribute should have been updated
Failed asserting that null is not equal to null.`

kernel/classes/datatypes/ezuser/ezuser.php
```diff
@@ -225,15 +225,24 @@ static function create( $contentObjectID )
         return new eZUser( $row );
     }
 
-    function store( $fieldFilters = null )
+    /**
+     * Only stores the entry if it has a Login value
+     *
+     * @param null $fieldFilters
```
glye:
```suggestion
     * @param mixed|null $fieldFilters
```
Not much point in a param that is always null :)

tests/tests/kernel/datatypes/ezuser/ezusertype_regression.php
```diff
@@ -110,8 +115,9 @@ public function testFromStringHandlesDisabledAccount()
         $userAccount = $dataMap['user_account']->fromString(
             join( '|', array( $tmpLogin, $tmpEmail, self::USER_PASSWORD_HASH, self::USER_PASSWORD_HASH_ID, 0 ) )
         );
-        $userAccount->store();
 
+        $dataMap['user_account']->setContent( $userAccount );
+        $dataMap['user_account']->store();
```
glye: Is this change needed, or cosmetic? :)

tests/tests/kernel/datatypes/ezuser/ezusertype_regression.php
```diff
@@ -188,11 +195,13 @@ public function testUpdatePasswordUpdatesSerializedData()
     {
         $userId = $this->userObject->attribute('id');
 
-        $passwordHash = $this->getSerializedPasswordHash($this->userObject);
+        $user = ezuser::fetch( $userId );
+        $passwordHash = $user->PasswordHash;
```
glye: Could you explain this change, please?

andrerom:
@pkamps can you accept / reject @glye's fixes?

kernel/classes/datatypes/ezuser/ezusertype.php
```diff
@@ -258,10 +258,22 @@ function storeObjectAttribute( $contentObjectAttribute )
             }
 
             // saving information in the object attribute data_text field to simulate a draft
-            $contentObjectAttribute->setAttribute( 'data_text', $this->serializeDraft( $user ) );
+            // only if the object version is a draft (status == 0)
+            if(
+                $user->Login &&
+                $contentObjectAttribute->attribute( 'object_version' )->attribute( 'status' ) == 0
```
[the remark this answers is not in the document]
pkamps: Hope you don't mind if I ignore it. Maybe it's a string and then it would hurt to be extra strict here.

kernel/user/ezuseroperationcollection.php
```diff
@@ -297,7 +297,7 @@ static public function activation( $userID, $userHash, $enableUser = false )
     }
 
    /**
-     * Change user password
+     * Change user password for an existing user
```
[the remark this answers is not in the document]
pkamps: I reverted my changes.

tests/tests/kernel/datatypes/ezuser/ezusertype_regression.php
pkamps: Those are required.

tests/tests/kernel/datatypes/ezuser/ezusertype_regression.php
pkamps: getSerializedPasswordHash was checking the 'data_text' value of the datatype. But the function eZUserOperationCollection::password is just updating the ezuser DB table. I changed the the test to more directly check the eZUserOperationCollection::password function.

The 'data_text' value in the datatype is only a temporarily place to store user data for draft versions. If an admin creates a new users, you don't want to have the draft data in the ezuser DB table right away.

kernel/classes/datatypes/ezuser/ezusertype.php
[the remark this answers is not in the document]
glye: We ought to be strict, strictly speaking ;) eZContentObjectVersion says it's an integer. But I see many cases where we aren't strict, so I guess this can pass if others agree.

glye: On the other hand, it makes sense here to use `eZContentObjectVersion::STATUS_DRAFT` instead of just `0`. Same in the comment above it.

glye: The comment could even be removed if we use the constant, since that makes the intention clear.

pkamps: That's a good point. I changed it now.

PR #1422, opened 2019-03-12
fix EZP-30293 : PHP Notice - Array to string

fwoldt:
legacy template navigator/google.tpl - PHP Notice - Array to string conversion
ignoring new view_parameters value _custom


andrerom:
Seems to duplicat of #1421 so closing this.

PR #1423, opened 2019-03-15
EZP-30316: Increase default minimum password length (legacy)

glye:
> Implement https://jira.ez.no/browse/EZP-30316

Our password requirement defaults should be safer. This change improves this for the default user content type. It does not affect existing users and their passwords, not until they decide to change password.

(If you need weaker passwords, in early dev maybe, you can easily change these defaults.)

- [x] Minimum password length: 10
- [x] Generated password length: 16

The same in eZ Platform: https://github.com/ezsystems/ezpublish-kernel/pull/2570


andrerom:
@emodric / @joekepley  / @crevillo Are you ok with this on 2017.12, or would you prefer 2019.03 for this given it is behaviour change?

emodric:
I'm okay for 2017.12 :+1:

glye:
Moved to master, after internal discussion.

glye:
Added upgrade doc.

glye:
@ezsystems/qa-team Please test. Lower priority than https://github.com/ezsystems/ezpublish-kernel/pull/2570

PR #1424, opened 2019-03-22
[WIP] PHP 7 compatibility issues

vidarl:
Various PHP7 changes.
For instance : PHP7.x left-right-order evaluation, ref. http://php.net/manual/en/migration70.incompatible.php


vidarl:
Included in https://github.com/ezsystems/ezpublish-legacy/pull/1428

PR #1427, opened 2019-04-01
remove hardcoded sortBy to read its value either from ini or post var

carlosmauri:
Hi,

I've removed the hardcoded sort by logic here: 
`extension/ezjscore/classes/ezjscserverfunctionsjs.php`
to be able to read either ini setting or post variable.

Now we can either use an ini setting to define the sort order we want to use globally or a new POST variable through the ajax call.
If the post variable is set, it will override the ini value.

If none of the values are set or the ezfind extension is OFF, there's a fallback that sets the $sortBy variable with the same "hardcoded" value.

This PR needs to be combined with this one:
https://github.com/ezsystems/ezfind/pull/220

Thanks!

### DIFF
--- a/extension/ezjscore/classes/ezjscserverfunctionsjs.php
+++ b/extension/ezjscore/classes/ezjscserverfunctionsjs.php
@@ -438,11 +438,28 @@ YUI( YUI3_config ).add('io-ez', function( Y )
         if ( self::hasPostValue( $http, 'EncodingDataMap' ) )
             $encodeParams['dataMap'] = $http->postVariable( 'EncodingDataMap' );
 
+	$ezFindIni = eZINI::instance( "ezfind.ini" );
+        if ( $ezFindIni->BlockValues[ 'SearchSettings' ] )
+        {
+            $sortBy = $ezFindIni->variable( "SearchSettings", "SortBy" );
+        }
+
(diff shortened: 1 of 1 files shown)


andrerom:
We don't usually place ezfind logic in legacy kernel.
If the fix is not already there, consider opening it there instead: https://github.com/ezsystems/ezfind/blob/master/classes/ezfindservercallfunctions.php#L16

Feel free to reopen if you think this is still relevant for kernel.

peterkeung:
The ezjscore search function already had eZ Find specific logic (see the "SortBy" param), and the kernel doesn't use that deprecated eZ Find search call.

andrerom:
Indeed you are right.
WDYT @emodric / @gggeek ?

gggeek:
Not a fan of this PR tbh. ezfind "knows" about ejscore. circualr deps are bad. ergo ezjscore should not "know" about ezfind (read an ini value from it).

I'd rather have ezjscore pass "NULL sort" to the search engine, and have the search engine determine what the default sorting is, or instead add SearchSettings/sortBy to kernel settings

gggeek:
Also, I would be happy as well with only handling here the case 'if received sort_by param via Post, honour it', and leave it up to the code that displays the search-page template and search link how to manage a default value...

PR #1390, opened 2018-09-20
PHP 5.3, 5.5, 5.6 and 7.0 deprecation fixes

pbek:
Fixes all that can be fixed of:

```
 PHP | File:Line                                                                                          |             Type | Issue
 5.3 | /extension/ezoe/ezxmltext/handlers/input/ezoexmlinput.php:661                                      | function_usage   | Function usage sectionLevel() (@call_with_passing_by_reference) is deprecated. 
 5.3 | /extension/ezoe/ezxmltext/handlers/input/ezoexmlinput.php:777                                      | function_usage   | Function usage inputSectionXML() (@call_with_passing_by_reference) is deprecated. 
 5.3 | /extension/ezoe/ezxmltext/handlers/input/ezoexmlinput.php:898                                      | function_usage   | Function usage inputListXML() (@call_with_passing_by_reference) is deprecated. 
 5.3 | /extension/ezoe/ezxmltext/handlers/input/ezoexmlinput.php:936                                      | function_usage   | Function usage inputTdXML() (@call_with_passing_by_reference) is deprecated. 
 5.3 | /extension/ezoe/ezxmltext/handlers/input/ezoexmlinput.php:965                                      | function_usage   | Function usage inputParagraphXML() (@call_with_passing_by_reference) is deprecated. 
 5.3 | /extension/ezoe/ezxmltext/handlers/input/ezoexmlinput.php:1049                                     | function_usage   | Function usage inputTagXML() (@call_with_passing_by_reference) is deprecated. 
 5.3 | /kernel/classes/datatypes/ezxmltext/ezxmloutputhandler.php:139                                     | function_usage   | Function usage viewTemplateSuffix() (@call_with_passing_by_reference) is deprecated. 
 5.3 | /lib/ezfile/classes/ezgzipzlibcompressionhandler.php:57                                            | function         | Function dl() is deprecated. 
 5.3 | /lib/ezpdf/classes/class.ezpdftable.php:1519                                                       | function_usage   | Function usage fixWhitespace() (@call_with_passing_by_reference) is deprecated. 
 5.3 | /lib/ezpdf/classes/class.pdf.php:2018                                                              | function         | Function set_magic_quotes_runtime() is deprecated. 
 5.3 | /lib/ezpdf/classes/class.pdf.php:2020                                                              | function         | Function set_magic_quotes_runtime() is deprecated. 
[...]

PR #1428, opened 2019-04-02
PHP 5.3, 5.5, 5.6 and 7.0 deprecation fixes

vidarl:
Continuation of https://github.com/ezsystems/ezpublish-legacy/pull/1390

### DIFF
--- a/extension/ezoe/modules/ezoe/classes/SpellChecker.php
+++ b/extension/ezoe/modules/ezoe/classes/SpellChecker.php
@@ -12,7 +12,7 @@ class SpellChecker {
 	 *
 	 * @param $config Configuration name/value array.
 	 */
-	function SpellChecker(&$config) {
+	function __construct(&$config) {
 		$this->_config = $config;
 	}
 
--- a/extension/ezoe/modules/ezoe/classes/SpellChecker.php
+++ b/extension/ezoe/modules/ezoe/classes/SpellChecker.php
@@ -3,7 +3,7 @@
  * $Id: editor_plugin_src.js 201 2007-02-12 15:56:56Z spocke $
  *
  * @author Moxiecode
- * @copyright Copyright © 2004-2007, Moxiecode Systems AB, All rights reserved.
+ * @copyright Copyright Â© 2004-2007, Moxiecode Systems AB, All rights reserved.
  */
 
 class SpellChecker {
@@ -16,13 +16,14 @@ class SpellChecker {
 		$this->_config = $config;
 	}
 
-    /**
-     * @deprecated Use SpellChecker::__construct() instead
-     * @param $config
-     */
+	/**
+	 * @deprecated Use SpellChecker::__construct() instead
+	 * @param $config
+	 */
 	function SpellChecker(&$config) {
 		self::__construct($config);
-    }
+	}
+
 	/**
 	 * Simple loopback function everything that gets in will be send back.
 	 *
--- a/extension/ezoe/modules/ezoe/classes/SpellChecker.php
+++ b/extension/ezoe/modules/ezoe/classes/SpellChecker.php
@@ -16,6 +16,13 @@ class SpellChecker {
 		$this->_config = $config;
 	}
 
+    /**
+     * @deprecated Use SpellChecker::__construct() instead
+     * @param $config
+     */
+	function SpellChecker(&$config) {
+		self::__construct($config);
+    }
 	/**
 	 * Simple loopback function everything that gets in will be send back.
 	 *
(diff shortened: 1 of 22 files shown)


gggeek:
Good, but misses compatibility constructors for the ones which have been renamed to __construct

lib/ezfile/classes/ezgzipzlibcompressionhandler.php
```diff
@@ -51,11 +51,15 @@ function compressionLevel()
     static function isAvailable()
     {
         $extensionName = 'zlib';
+
+        // dl() is removed in PHP 7.0, see https://secure.php.net/manual/en/function.dl.php
+        /*
```
andrerom: Please rather remove all commented out code, feel free to clean up surrounding commented out code as well _(ref some of the `set_magic_quotes_runtime ` changes)_

lib/ezfile/classes/ezgzipzlibcompressionhandler.php
```diff
@@ -50,13 +50,7 @@ function compressionLevel()
...
     static function isAvailable()
     {
-        $extensionName = 'zlib';
-        if ( !extension_loaded( $extensionName ) )
-        {
-            $dlExtension = ( eZSys::osType() == 'win32' ) ? '.dll' : '.so';
-            @dl( $extensionName . $dlExtension );
-        }
-        return extension_loaded( $extensionName );
+        return function_exists('gzcompress');
```
vidarl: Okay just to check for one zlib function? or should I check for all of those used in class?

vidarl: (checking existence of function instead of extension due to feedback from @gggeek to support polyfills )

vidarl:
> Good, but misses compatibility constructors for the ones which have been renamed to __construct
Fixed

lib/ezfile/classes/ezgzipzlibcompressionhandler.php
vidarl: Replaced code with `function_exists('gzcompress')`

Also removed commented `set_magic_quotes_runtime` stuff

extension/ezoe/modules/ezoe/classes/SpellChecker.php
```diff
@@ -12,10 +12,17 @@ class SpellChecker {
...
+	function __construct(&$config) {
 		$this->_config = $config;
 	}
 
+    /**
+     * @deprecated Use SpellChecker::__construct() instead
+     * @param $config
+     */
+	function SpellChecker(&$config) {
+		self::__construct($config);
```
andrerom: CS (same for all others it seems)

andrerom: hmm, this is already issue in the file from before from tinymce source

then ignore, but the docblock seems off here compared to other methods

lib/ezfile/classes/ezgzipzlibcompressionhandler.php
andrerom: @gggeek any opinion on this? I guess if they all are part of same extension, checking one should be ok'ish. The prior code did not take issue of certain functions being disabled manually into account anyway.

lib/ezutils/classes/ezsendmailtransport.php
```diff
@@ -31,19 +31,15 @@ function sendMail( eZMail $mail )
...
-        if ( !$isSafeMode and
-             $emailSender )
+        if ( $emailSender )
             $sendmailOptions .= ' -f'. $emailSender;
 
-        if ( $isSafeMode and
-             $emailSender and
+        if ( $emailSender and
              $mail->sender() == false )
             $mail->setSenderText( $emailSender );
```
andrerom: should this still be here?
(seems emailSender is set above for non safe mode, so seems it's set two times now in different ways)

lib/ezfile/classes/ezgzipzlibcompressionhandler.php
gggeek: I'd say: we use up to 3 functions: check for all of them. More than that: check just one...

I cant come up with a 'perfect' plan as each strategy has drawbacks:
- check only for 1 function: we incur the risk that others are not poly-filled
- check for all of them: makes code slow and unreadable
- check for just a couple: makes sense if those are the most-frequently used, eg. they cover 80% of usages by the current class. but also worst of both worlds...

andrerom:
From my side this is almost ready, only open topic is the one in `lib/ezutils/classes/ezsendmailtransport.php` which seems to indicate sender is set twice now in different ways.

lib/ezutils/classes/ezsendmailtransport.php
vidarl: Fixed in f4fb830

PR #1429, opened 2019-04-15
EZP-30416: INI settings don't support loose typing

paulortvalle:
Update the set definition to an empty array.
Instead of false, if not set, it uses the sorting defined for each node (from the Ordering tab).

### DIFF
--- a/settings/contentstructuremenu.ini
+++ b/settings/contentstructuremenu.ini
@@ -36,9 +36,8 @@ MaxNodes=150
 #    SortBy[]
 #    SortBy[]=name/ascending
 #    SortBy[]=published/descending
-# if set to "false" - default sorting will be implemented
-# (which is defined for each node)
-SortBy=false
+# if not set, it uses the sorting defined for each node (from the Ordering tab)
+SortBy=[]
 
 # enabled/disabled
 ToolTips=enabled
--- a/settings/contentstructuremenu.ini
+++ b/settings/contentstructuremenu.ini
@@ -36,7 +36,7 @@ MaxNodes=150
 #    SortBy[]
 #    SortBy[]=name/ascending
 #    SortBy[]=published/descending
-# if not set, it uses the sorting defined for each node (from the Ordering tab)
+# if an empty array, it uses the sorting defined for each node instead (from the Ordering tab)
 SortBy[]
 
 # enabled/disabled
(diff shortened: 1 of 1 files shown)


settings/contentstructuremenu.ini
```diff
@@ -36,9 +36,8 @@ MaxNodes=150
 #    SortBy[]
 #    SortBy[]=name/ascending
 #    SortBy[]=published/descending
-# if set to "false" - default sorting will be implemented
-# (which is defined for each node)
-SortBy=false
+# if not set, it uses the sorting defined for each node (from the Ordering tab)
+SortBy=[]
```
andrerom: That is not a INI array value.
Emptying and setting the value as array is as documented `SortBy[]`.
Appending is done by means of `SortBy[]=name/ascending`

paulortvalle: Thank you @andrerom!

andrerom:
@paulortvalle I'm not entirely sure why you want the default setting to change here, is there a bug somewhere? If so what do you exactly do to trigger it and what is your expected behaviour?

So maybe add some explanation and a link to the jira issue so it's possible to connect the dots on why you propose this change ;)

paulortvalle:
Hello @andrerom !

Thank you for you fast answer.
The link for the ticket on Jira is: https://jira.ez.no/browse/EZP-30416

I will explain how to get this error:
if you use that setting with an array and try to view the contentstructuremenu.ini settings in Setup > Ini settings, you get this error:

An unexpected error has occurred. Please contact the webmaster.
[] operator not supported for strings in /var/www/marigold/ezpublish_legacy/lib/ezutils/classes/ezini.php on line 868

So, initialize the setting with false will break the file view interface on admin interface.

Please, let me know if you need any other information. I'm happy to help.

settings/contentstructuremenu.ini
```diff
@@ -36,9 +36,8 @@ MaxNodes=150
 #    SortBy[]
 #    SortBy[]=name/ascending
 #    SortBy[]=published/descending
-# if set to "false" - default sorting will be implemented
-# (which is defined for each node)
-SortBy=false
+# if not set, it uses the sorting defined for each node (from the Ordering tab)
```
andrerom:
```suggestion
# if empty array, it uses the sorting defined for each node (from the Ordering tab)
```

peterkeung:
This has been running fine on the projects where we changed this. Pinging @paulortvalle to rebase.

paulortvalle:
@andrerom 
Ready for you!

settings/contentstructuremenu.ini
andrerom: @paulortvalle Are you ok with this change above? I just want to make sure we are clear clear here ;)

mateuszbieniek (approved):

PR #1430, opened 2019-05-03
[2019.03] Some fixes for MySQL8 and PHP7.3

andrerom:
This does not imply full MYSQL8 support, but at least it makes it possible to login.


andrerom:
yes, but it at least won't fail for any other hash type like it does now.

PR #1420, opened 2019-02-22
EZP-30161 Update ezjscore token variable identifier

thiagocamposviana:
https://jira.ez.no/browse/EZP-30161

New versions should use _token and not ezxform_token_js

PR #1431, opened 2019-05-03
EZP-30161: Handle form token in both custom and also for historical reasons default value

andrerom:
Since 5.0 ezxFormToken code has been operating in two modes:
- Default pure legacy mode using `ezxform_token` as form name
- Symfony mode using `_token` as form name _(if symfony settings are in default state)_

However quite some legacy code hard codes the default name, at least for javacript usage.
In 5.x this inconsistency was solved by setting Symfony to use `field_name: ezxform_token`

For Platform + legacy bridge that is no longer the case, and neither do we want to change the field_name by default. So this changes the logic so that ezxFormToken is cable of handling both at the same time.

Closes #1420

@thiagocamposviana Could you have a look and test if this solves your issue?

### DIFF
--- a/extension/ezformtoken/event/ezxformtoken.php
+++ b/extension/ezformtoken/event/ezxformtoken.php
@@ -38,7 +38,7 @@ class ezxFormToken
     static protected $intention = 'legacy';
 
     /**
-     * @var string
+     * @var string Custom Form field, by default set to system default form field (self::FORM_FIELD).
      */
     static protected $formField = self::FORM_FIELD;
 
@@ -90,6 +90,8 @@ class ezxFormToken
     }
 
     /**
+     * Get the custom form field.
+     *
      * @return string
      */
     static protected function getFormField()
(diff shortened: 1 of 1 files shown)


extension/ezformtoken/event/ezxformtoken.php
```diff
@@ -188,19 +197,22 @@ static public function output( $templateResult, $filterForms = true )
...
 
         $token = self::getToken();
-        $field = self::getFormField();
+        $customfield = self::getFormField();
+        $defaultField = self::FORM_FIELD;
         $replaceKey = self::REPLACE_KEY;
 
         eZDebugSetting::writeDebug( 'ezformtoken', 'Output protected (all forms will be modified)', __METHOD__ );
 
+        // Inject token for programmatically use (also system default for historical reasons)
```
glye:
```suggestion
        // Inject token for programmatical use (also system default for historical reasons)
```

thiagocamposviana:
Just want to confirm it works nicely!

PR #1434, opened 2019-05-28
Popup for the custom tag sub/sup fixed

pkamps:
The sub (subscript) and sup (superscript) tags work, but there is a UI issue.

Hitting "Ok" once adds the tag, but does not close the dialog. The dialog only closes upon clicking "Cancel."

While the dialog is open, each click of "Ok" after the first inserts the highlighted text at the beginning of the paragraph, in the sub or sup style.

Hitting "Cancel" is the only way out, which makes the necessary workflow:

Select text
Select Custom tags
Select "Sup"
Click OK
Click Cancel

**Reproduce / Testing instructions**

Select text
Select Custom tags
Select "Sup"
Click OK => Dialog remains open
Click OL => Duplicate text inserted at beginning of paragraph.

Tested with: Mac Chrome.

### DIFF
--- a/extension/ezoe/design/standard/templates/ezoe/tag_custom.tpl
+++ b/extension/ezoe/design/standard/templates/ezoe/tag_custom.tpl
@@ -101,6 +101,11 @@ tinyMCEPopup.onInit.add( eZOEPopupUtils.BIND( eZOEPopupUtils.init, window, {
     },
     onTagGenerated:  function( el, ed, args )
     {
+        if( el === null )
+        {
+            return;
+        }
+
         // append a paragraph if user just inserted a custom tag in editor and it's the last tag
         var edBody = el.parentNode, doc = ed.getDoc(), temp = el;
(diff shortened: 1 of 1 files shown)


extension/ezoe/design/standard/templates/ezoe/tag_custom.tpl
```diff
@@ -101,6 +101,11 @@ tinyMCEPopup.onInit.add( eZOEPopupUtils.BIND( eZOEPopupUtils.init, window, {
     },
     onTagGenerated:  function( el, ed, args )
     {
+        if( el === null )
```
andrerom:
```suggestion
        if ( el === null )
```

andrerom:
@pkamps Do you need this on previous versions also _(not just 2019.03)_ ? If so are you able to rebase this?

pkamps:
Not needed in any particular version. Just sharing my effort on the legacy code.

PR #1435, opened 2019-07-03
Fix EZP-30718: Allow RerunURL to be used on hidden pages

peterkeung:
https://jira.ez.no/browse/EZP-30718

The Rerun URL instructions get trapped in the view cache (for hidden pages) because the redundant check for the module exit status can never pass once a page is cached.

The check against "rerun_uri" in the module parameters is sufficient.

### DIFF
--- a/kernel/layout/set.php
+++ b/kernel/layout/set.php
@@ -17,7 +17,6 @@ foreach ( $Params['UserParameters'] as $key => $param )
 
 $Result = array();
 $Result['content'] = '';
-$Result['rerun_uri'] = '/' . implode( '/', array_splice( $Params['Parameters'], 1 ) ) . $userParamString;
 
 $layoutINI = eZINI::instance( 'layout.ini' );
 $i18nINI = eZINI::instance( 'i18n.ini' );
@@ -53,6 +52,7 @@ if ( $layoutINI->hasGroup( $LayoutStyle ) )
         }
     }
 
+    $Result['rerun_uri'] = '/' . implode( '/', array_splice( $Params['Parameters'], 1 ) ) . $userParamString;
     $Module->setExitStatus( eZModule::STATUS_RERUN );
 }
 else
--- a/kernel/private/classes/ezpkernelweb.php
+++ b/kernel/private/classes/ezpkernelweb.php
@@ -843,15 +843,14 @@ class ezpKernelWeb implements ezpWebBasedKernelHandler
                 $moduleResult = $this->module->handleError( eZError::KERNEL_MODULE_DISABLED, 'kernel', array( 'check' => $moduleCheck ) );
             }
             $this->siteBasics['module-run-required'] = false;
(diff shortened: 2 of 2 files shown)


kernel/layout/set.php
```diff
@@ -17,7 +17,7 @@
 
 $Result = array();
 $Result['content'] = '';
-$Result['rerun_uri'] = '/' . implode( '/', array_splice( $Params['Parameters'], 1 ) ) . $userParamString;
+$rerunURL = '/' . implode( '/', array_splice( $Params['Parameters'], 1 ) ) . $userParamString;
```
andrerom: what is the point of this change?

peterkeung: Prevent BC break: only return "rerun_uri" in the module result if the layout exists, since the PR makes it so the kernel only checks "rerun_uri" rather than the module's exit status. Alternatively we could just move this code further down to where $rerunURL is used.

andrerom: > Alternatively we could just move this code further down to where $rerunURL is used.

👍 that would be cleaner / easier to read.

kernel/private/classes/ezpkernelweb.php
```diff
@@ -843,15 +843,10 @@ protected function dispatchLoop()
...
-                if ( isset( $moduleResult['rerun_uri'] ) )
-                {
-                    $this->uri = eZURI::instance( $moduleResult['rerun_uri'] );
-                    $this->siteBasics['module-run-required'] = true;
-                }
-                else
-                    eZDebug::writeError( 'No rerun URI specified, cannot continue', 'index.php' );
+                $this->uri = eZURI::instance( $moduleResult['rerun_uri'] );
+                $this->siteBasics['module-run-required'] = true;
             }
```
andrerom: Maybe also add something like this to retain the info on inconsistent status and missing rerun uri.

```suggestion
            }
            else if ( $this->module->exitStatus() == eZModule::STATUS_RERUN )
            {
                eZDebug::writeError( 'No rerun URI specified on eZModule::STATUS_RERUN , cannot set URI', 'index.php' );
            }
```

kernel/private/classes/ezpkernelweb.php
```diff
@@ -843,15 +843,10 @@ protected function dispatchLoop()
...
```
andrerom: Maybe also add something like this to retain the info on inconsistent status and missing rerun uri.

```suggestion
            else if ( $this->module->exitStatus() == eZModule::STATUS_RERUN )
            {
                eZDebug::writeError( 'No rerun URI specified on eZModule::STATUS_RERUN , cannot set URI', 'index.php' );
            }
```

kernel/layout/set.php
peterkeung: Done, thanks!

kernel/private/classes/ezpkernelweb.php
peterkeung: Done, thanks!

PR #1437, opened 2019-07-19
Fix EZP-30767: Allow the developer to override the page to load on a partial URL alias path

peterkeung:
https://jira.ez.no/browse/EZP-30767

It is not desirable to load the homepage on what should otherwise be a 404 page.

To prevent a BC break, simply allow this behavior to be overridden.


mateuszbieniek:
Looks OK for me, thank you for your contribution.

mateuszbieniek:
Note: the failing test is not related to this PR.

PR #1443, opened 2019-08-13
Updated instances of count() that would cause a warning in 7.2

steepmountain:
Logic updated after the following example to give a reasonable approximation of how count() worked on scalar objects previous to PHP 7.2

https://stackoverflow.com/questions/49662029/count-emitting-an-e-warning

### DIFF
--- a/kernel/classes/ezcontentobjecttreenode.php
+++ b/kernel/classes/ezcontentobjecttreenode.php
@@ -4168,7 +4168,7 @@ class eZContentObjectTreeNode extends eZPersistentObject
 
         $nodePath = null;
         $nodeDepth = 0;
-        if (((is_array($node) || $node instanceof Countable) && count($node) != 0) || $node != null) {
+        if ( is_countable( $node ) && count( $node ) != 0 ) {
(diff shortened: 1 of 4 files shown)


kernel/classes/ezcontentobjecttreenode.php
```diff
@@ -4168,12 +4168,10 @@ function subtreeSoleNodeCount( $params = array() )
 
         $nodePath = null;
         $nodeDepth = 0;
-        if ( count( $node ) != 0 )
-        {
+        if (((is_array($node) || $node instanceof Countable) && count($node) != 0) || $node != null) {
```
vidarl: Since we require `symfony/polyfill-php73`, you may simplify these using is_countable(), see https://github.com/ezsystems/ezpublish-legacy/commit/6321a92b4b9a58ef0965415a369dba3a71e79d24#diff-6af1184676109776bea24cee6c02acb7 for more info

steepmountain: Thanks for the info, I wasn't aware. Fixed the code to implement `is_countable` instead.

PR #1444, opened 2019-08-19
Fix section update on move

SerheyDolgushev:
Right now, moved content section is updated, if:

- it is the main location (https://github.com/ezsystems/ezpublish-legacy/blob/2019.03/kernel/classes/ezcontentobjecttreenodeoperations.php#L81)
- AND moved content section is not the same as new parents (https://github.com/ezsystems/ezpublish-legacy/blob/2019.03/kernel/classes/ezcontentobjecttreenodeoperations.php#L86)
- AND moved content section is the same as its old parent content section (https://github.com/ezsystems/ezpublish-legacy/blob/2019.03/kernel/classes/ezcontentobjecttreenodeoperations.php#L91)

This PR disables the last condition.

### DIFF
--- a/kernel/classes/ezcontentobjecttreenodeoperations.php
+++ b/kernel/classes/ezcontentobjecttreenodeoperations.php
@@ -87,8 +87,7 @@ class eZContentObjectTreeNodeOperations
                 {
 
                     eZContentObjectTreeNode::assignSectionToSubTree( $newNode->attribute( 'main_node_id' ),
-                                                                     $newParentObject->attribute( 'section_id' ),
-                                                                     $oldParentObject->attribute( 'section_id' ) );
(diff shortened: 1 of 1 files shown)


pkamps:
Could you add steps how to reproduce the issue? Something like a description what to do in the admin interface in order to run into the issue.

SerheyDolgushev:
@pkamps can you please check out https://github.com/ezsystems/ezpublish-kernel/pull/2741 ?

vidarl:
Based on https://github.com/ezsystems/ezpublish-kernel/pull/2741#issuecomment-523408828, shouldn't this one be closed too? Or am I missing anything?

andrerom:
Yes, this can be closed. Intention is rather to align Platform with legacy behaviour by default.

PR #1446, opened 2019-11-01
[Travis] Add testing for PHP 7.2 and 7.3

andrerom:


gggeek:
Not sure if it's relevant here - possibly the ezp legacy test suite is tightly coupled to phpunit version 4, but I had good success running unit tests with different phpunit versions to achieve compatibility with php 5.6 => 7.3 using version constraint "phpunit/phpunit": "^5.7 || ^7.5" `

andrerom:
> I had good success running unit tests with different phpunit versions to achieve compatibility with php 5.6 => 7.3 using version constraint "phpunit/phpunit": "^5.7 || ^7.5" `

Was that on legacy? that would be relevant, or are you talking about another package?

gggeek:
^^ nopes, other project

andrerom:
ok, well seems travis is ok with this so we just got  testing up and running.

There are deprecation messages coming from phpunit code over `each()` usage, but they don't interfere with the test run and results so I deemed this good enough for now.

----

If anyone wants to take this further to fix that, this might be relevant:
I did check how much work it would be to use `symfony/phpunit-bridge`, but we will need to change to use xml suites and several other changes in order to be able to run tests via it from what I could see, this will affect extensions with tests as well.

PR #1447, opened 2019-11-20
Fix: Missing ob_end_clean()

blankse:
Without this line in commands the output is buffered till the script is ended.

### DIFF
--- a/kernel/classes/datatypes/ezbinaryfile/plugins/ezpdfparser.php
+++ b/kernel/classes/datatypes/ezbinaryfile/plugins/ezpdfparser.php
@@ -39,6 +39,7 @@ class eZPDFParser
         // fill the buffer with the old values
         ob_start();
         print( $buffer );
+        ob_end_clean();
 
         return $metaData;
     }


gggeek:
I am not sure about your fix tbh, even thought the code in that handler is weird.. it seems to assume that the code _is always running_ inside an ob_start() section.

A better fix is, imho:
a) replace usage of 'passthru' above with 'exec'
b) remove all ob_ s**t

blankse:
Yes I closed it now.

It should be like
```
class eZPDFParser
{
    function parseFile( $fileName )
    {
        $binaryINI = eZINI::instance( 'binaryfile.ini' );

        $textExtractionTool = $binaryINI->variable( 'PDFHandlerSettings', 'TextExtractionTool' );
        $obLevel = ob_get_level();
        // save the buffer contents
        $buffer = ob_get_contents();

        if ( $obLevel && $buffer )
        {
            ob_end_clean();
        }

        // fetch the module printout
        ob_start();
        passthru( "$textExtractionTool $fileName" );
        $metaData = ob_get_contents();
        ob_end_clean();

        if ($obLevel) {
            // fill the buffer with the old values
            ob_start();
            print( $buffer );
        }

        return $metaData;
    }
}
```

gggeek:
that seems good too

PR #1448, opened 2019-11-21
Make autoloads ignore anonymous classes

Pr0metheus42:


emodric:
Having the following notice now:

```
Scanning for PHP-files.

Scan complete. Found 497 PHP files.

PHP Notice:  Undefined offset: -1 in /var/www/html/netgensite/ezpublish_legacy/kernel/private/classes/ezautoloadgenerator.php on line 646
PHP Notice:  Undefined offset: -1 in /var/www/html/netgensite/ezpublish_legacy/kernel/private/classes/ezautoloadgenerator.php on line 646
PHP Notice:  Undefined offset: -1 in /var/www/html/netgensite/ezpublish_legacy/kernel/private/classes/ezautoloadgenerator.php on line 646

Found 183 classes, added 183 of them to the autoload array.
```

andrerom:
So need to check existence of T_NEW constant then?
Would one of you be able to open PR?

emodric:
@andrerom https://github.com/ezsystems/ezpublish-legacy/pull/1450

PR #1450, opened 2019-12-18
Fix notice when checking for anonymous  classes in autoload generator

emodric:


andrerom:
Fixed in https://github.com/ezsystems/ezpublish-legacy/releases/tag/v2019.03.4

PR #1451, opened 2019-12-18
Reword the php7 BC doc and fix spelling errors

gggeek:


andrerom:
Nice, but could you rebase / re-applay on 2019.03 branch? _(it's the default branch since there might never be any new major release, so it's the one most people brose and use)_

gggeek:
Created #1452

PR #1453, opened 2019-12-30
Show images after used url_prefix

anijra:
After using AWS S3 as the source of external images, they do not appear in the rich text editor and in the modal of inserting images, this pull request will be resolved this problem.

### DIFF
--- a/extension/ezjscore/classes/ezjscajaxcontent.php
+++ b/extension/ezjscore/classes/ezjscajaxcontent.php
@@ -478,7 +478,8 @@ class ezjscAjaxContent
                                     (string)$element, 'UTF-8'
                                 );
                             }
-                            if($key === 'url') {
+                            if( $key === 'url' )
+                            {
                                 eZURI::transformURI( $element, true );
                             }
                         }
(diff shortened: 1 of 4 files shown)


anijra:
Hi, before the patch the external images do not receive the same host name just "/", and by the patch it receives the AWS host.
For the path, in both they are correct.

glye:
@anijra Thanks. I'm wondering if image links on non-AWS sites also will include the hostname with this fix. It should work well in either case.

emodric:
Hi!

It seems this issue causes a regression where image added to a XML text field ends up with a starting double slash in the source code:

image

This ultimately causes Chrome to request the image with `http://var/...` URI, which causes the image not to be displayed in the editor.

/cc @iherak

glye:
@emodric This is in cases *not* using AWS S3, or what?

emodric:
Yep, normal local install, no DFS or nothing.

glye:
So the quickest solution is to revert this. Better to break AWS, than everything else (at least non-dfs). And then do a better-tested aws fix later. 2nd opinion... @bdunogier maybe?

PR #1454, opened 2020-02-20
PHP7.2 fix: Added array check before count.

StianTorjussen:
The call to get a post variable `$http->postVariable( $postVariableName )` sometimes returned `null`.
Resulting in `count(null)`, which triggers a warning.

Ref. phpdoc changelog for count: 
[https://www.php.net/manual/en/function.count.php#refsect1-function.count-changelog](https://www.php.net/manual/en/function.count.php#refsect1-function.count-changelog)

### DIFF
--- a/kernel/classes/datatypes/ezobjectrelationlist/ezobjectrelationlisttype.php
+++ b/kernel/classes/datatypes/ezobjectrelationlist/ezobjectrelationlisttype.php
@@ -76,7 +76,7 @@ class eZObjectRelationListType extends eZDataType
             // If in browse mode and relations have been added using the search field
             // items are stored in the post variable
             if (
-                is_array( $http->postVariable( $postVariableName ) )
+                $http->hasPostVariable( $postVariableName )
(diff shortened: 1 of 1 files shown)


kernel/classes/datatypes/ezobjectrelationlist/ezobjectrelationlisttype.php
```diff
@@ -76,7 +76,8 @@ function validateObjectAttributeHTTPInput( $http, $base, $contentObjectAttribute
             // If in browse mode and relations have been added using the search field
             // items are stored in the post variable
             if (
-                $http->postVariable( $postVariableName ) != array( "no_relation" )
+                is_array( $http->postVariable( $postVariableName ) )
```
andrerom: If it's null I guess it's not set, so then more correct would be to make sure we check if it's here:
```suggestion
                $http->hasPostVariable( $postVariableName )
```

StianTorjussen: Yes, that seems reasonable.

PR #1455, opened 2020-05-12
store can not be called here, because it clears object's data_map

SerheyDolgushev:
https://github.com/ezsystems/ezpublish-legacy/pull/819


SerheyDolgushev:
@glye, please feel free to merge this instead of https://github.com/ezsystems/ezpublish-legacy/pull/819

glye:
Merged (with an edited commit message). Thanks, and sorry about the 7 year delay 😊

PR #1457, opened 2020-05-22
Fix handling of headings in table header cells

hknezevic:
If you add heading element inside the regular table cell (td), it will be correctly applied, stored and displayed on the frontend.
If you try to do the same inside table header cell (th), after you store the changes, the heading will get moved from the header cell and placed below the table.

This PR fixes the incorrect behaviour.


pkamps:
I was able to confirm the behaviour. I had to disable the editor in order to change the cells to <th>.

PR #1458, opened 2020-06-04
Pr819

pkamps:
Upstream pull request: https://github.com/ezsystems/ezpublish-legacy/pull/1455


pkamps:
doh, sorry about that. Shouldn't work so late.

PR #1459, opened 2020-06-12
Send unhandled exception error messages to `error_log()`.

omh:
Before PHP 7 fatal errors was written to whatever PHP's `error_log` was set to. In PHP 7 most errors are exceptions, so eZ ends up swallowing everything, including fatal errors, in it's default exception handler and logging to its own `var/log/error.log` file. This file is, by default, is *heavily* rotated after only 250KB. This means it's very easy to lose important error messages.

### DIFF
--- a/lib/ezutils/classes/ezexecution.php
+++ b/lib/ezutils/classes/ezexecution.php
@@ -202,7 +202,9 @@ class eZExecution
             }
         }
 
-        eZLog::write( 'Unexpected error, the message was : ' . $e->getMessage() . ' in ' . $e->getFile() . ' on line ' . $e->getLine(), 'error.log' );
+        $msg = 'Unexpected error, the message was : ' . $e->getMessage() . ' in ' . $e->getFile() . ' on line ' . $e->getLine();
+        eZLog::write( $msg, 'error.log' );
+        error_log( $msg );
 
         eZExecution::cleanup();
         eZExecution::setCleanExit();


lib/ezutils/classes/ezexecution.php
```diff
@@ -202,7 +202,9 @@ static public function defaultExceptionHandler( $e )
             }
         }
 
-        eZLog::write( 'Unexpected error, the message was : ' . $e->getMessage() . ' in ' . $e->getFile() . ' on line ' . $e->getLine(), 'error.log' );
+        $msg = 'Unexpected error, the message was : ' . $e->getMessage() . ' in ' . $e->getFile() . ' on line ' . $e->getLine();
+        eZLog::write( msg, 'error.log' );
```
guillaumelecerf: misssing \$ before `msg`

guillaumelecerf (changes requested):

lib/ezutils/classes/ezexecution.php
omh: That's embarrassing 😱  Fixed now :)

guillaumelecerf (approved):

glye (approved):

andrerom:
@omh if possible rebase for 2019.03, as there are no longer any releases made from master.

pkamps:
I understand that you can easily miss a log message due to the heavy log rotation. Would it help to adjust the log rotation with following setting?

https://github.com/ezsystems/ezpublish-legacy/blob/2019.03/config.php-RECOMMENDED#L250

I don't like the idea to push the error message into 2 different channels (error_log and eZLog::write).

We can also consider to raise the max log file size: 250KB is really low. I don't see an advantage that the files are that small (other than less disk space usage):
https://github.com/mugoweb/ezpublish-legacy/pull/194

omh:
@pkamps what specifically is it that you do not like about pushing it to two different channels? Potential performance impact? (which I think should be negligible assuming your site isn't crashing _a lot_).

pkamps:
Hi,

sorry for the late reply.

The performance impact is probably very low.

All software solutions I worked with have different log channels for notice, warning, errors etc. 
But I haven't seen that software solutions typically duplicates each log message and sends them to 2 different endpoints (files in most cases or in your case error_log and eZLog::write).

If you really need to duplicate each message, it's probably cleaner to extend eZLog and allow it to send messages to 2 endpoints.

omh:
I'm no longer involved with ezpublish legacy so closing this.

PR #1460, opened 2020-08-28
Removing an eZOrder should also remove its eZOrderItems

gabrielfin:


andrerom (approved):

kernel/classes/ezorder.php
```diff
@@ -1353,6 +1353,7 @@ static function cleanupOrder( $orderID )
             $db = eZDB::instance();
             $db->begin();
             $db->query( "DELETE FROM ezorder where id='$orderID'" );
+            $db->query( "DELETE FROM ezorder_item WHERE order_id=$orderID" );
```
andrerom, line 1356: Looking at line below, should this have been:
```suggestion
            $db->query( "DELETE FROM ezorder_item WHERE order_id=$orderNr" );
```

??

gabrielfin:
I don't know why `ezorder_status_history` uses orderNr instead of orderID, (very confusing btw, since the column and attribute is called `order_id`) but as far as I can tell, `order_item` _does_ use orderID. Take a look [here](https://github.com/ezsystems/ezpublish-legacy/blob/3291c089e04d5698cd7657374af1cbb86c093c70/kernel/classes/ezorder.php#L949)

PR #1464, opened 2020-10-12
EZP-32018: Fixed empty ezdate field transformation

barw4:
JIRA issue: [EZP-32018](https://jira.ez.no/browse/EZP-32018)

Empty field value for `ezdate` fieldtype has been handled additionally.

### DIFF
--- a/lib/ezutils/classes/eztimestamp.php
+++ b/lib/ezutils/classes/eztimestamp.php
@@ -15,7 +15,9 @@ class eZTimestamp
     */
     public static function getUtcTimestampFromLocalTimestamp( $localTimestamp ) {
 
-        if (!$localTimestamp) return null;
+        if (!$localTimestamp) {
+            return null;
+        }
 
         $utcTimezone = new \DateTimeZone( 'UTC' );
         $localTimezone = new \DateTimeZone( date_default_timezone_get() );
@@ -32,7 +34,9 @@ class eZTimestamp
     */
     public static function getLocalTimestampFromUtcTimestamp( $utcTimestamp ) {
 
-        if (!$utcTimestamp) return null;
+        if (!$utcTimestamp) {
+            return null;
+        }
 
         $utcTimezone = new \DateTimeZone( 'UTC' );
         $localTimezone = new \DateTimeZone( date_default_timezone_get() );
--- a/lib/ezutils/classes/eztimestamp.php
+++ b/lib/ezutils/classes/eztimestamp.php
@@ -15,7 +15,8 @@ class eZTimestamp
     */
     public static function getUtcTimestampFromLocalTimestamp( $localTimestamp ) {
 
-        if (!$localTimestamp) {
+        if ( !$localTimestamp )
+        {
             return null;
         }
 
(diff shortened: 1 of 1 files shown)


konradoboza (approved):

lib/ezutils/classes/eztimestamp.php
```diff
@@ -14,6 +14,9 @@ class eZTimestamp
      \return a timestamp in UTC
     */
     public static function getUtcTimestampFromLocalTimestamp( $localTimestamp ) {
+
+        if (!$localTimestamp) return null;
```
glye, line 18: Not sure if even legacy CS allows brace-less ifs. Same for other change.
```suggestion
        if (!$localTimestamp)
        {
            return null;
        }
```

glye (approved):
Ok but not CS kosher, AFAICR.

lib/ezutils/classes/eztimestamp.php
```diff
@@ -14,6 +14,11 @@ class eZTimestamp
      \return a timestamp in UTC
     */
     public static function getUtcTimestampFromLocalTimestamp( $localTimestamp ) {
+
+        if (!$localTimestamp) {
```
andrerom, line 18:
```suggestion
        if ( !$localTimestamp ) {
```

legacy CS ;)

andrerom (approved):

lib/ezutils/classes/eztimestamp.php
```diff
@@ -14,6 +14,11 @@ class eZTimestamp
...
+            return null;
```
andrerom, line 19: Legacy CS:

```
if ( !$localTimestamp )
{
```

andrerom: So:
1. spacing inside parentheses
2. blocks always on newline

glye: 👆

barw4: Ah, wasn't aware of that, resolved in: https://github.com/ezsystems/ezpublish-legacy/pull/1464/commits/1226e4c9df174e5effdfb582f471cd56cf6a35bc

andrerom:
👍

glye (approved):

andrerom:
@barw4 Can you merge this, merge 2019.03 into master, and if your up for it also release v2019.03.6 (I can show you how).

barw4:
Merged into master: https://github.com/ezsystems/ezpublish-legacy/commit/10902ec20de7235c8b1423e5f1d2e3a8998b2b7e

pkamps:
It's unlikely, but what about if you actually want the timestamp to be '0'? That would return _NULL_ instead of 1.1.1970, right?

gggeek:
@pkamps you are right. I am pretty sure that there have been changes not too long ago in the ez5 kernel to deal with 0/null as default and 0 as valid datetime. I think here we should reproduce the same logic (might be even worth some unit testing)...

pkamps:
You're talking about this one? https://jira.ez.no/browse/EZP-16509

We fixed it in our fork - still an open pull request: https://github.com/mugoweb/ezpublish-legacy/pull/188

barw4:
@pkamps thank you for the catch, fixed here: https://github.com/ezsystems/ezpublish-legacy/pull/1465

PR #1466, opened 2020-11-04
EZP-32130: Set version of polyfill-php73 to 1.19.* for PHP comatibility >= 5.5.0

iegorka:
JIRA ticket https://issues.ibexa.co/browse/EZP-32130

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,7 +40,7 @@
         "ezsystems/ezprestapiprovider-ls": "~5.3",
         "ezsystems/ezstarrating-ls-extension": "~5.3",
         "ezsystems/ezwt-ls-extension": "~5.3",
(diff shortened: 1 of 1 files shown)


andrerom:
Why do you need this? composer is supposed to pick 1.19 for php versions not supported by 1.20. 

Unless you install via composer with things like `--ignore-platform-reqs`, or you do install on PHP 7 and then run on PHP 5. For both cases you should then set "platform" composer variable to PHP 5.5 if you need to on your side. See:
https://getcomposer.org/doc/06-config.md#platform

iegorka:
Hello,

I'm really sorry for this fictional problem. 
You're right with I did the installation on PHP 7.3 and then it didn't work on PHP 7.0

Thank you for your suggestions.

PR #1467, opened 2022-04-29
Update ezselectiontype.php to actually detect PHP array Input as empt…

davidsayre:
…y to enforce required


davidsayre:
error in code, fixing now

PR #1472, opened 2025-11-27
202511 702 session cookies samesite support

carlosmauri:
Hi @pkamps ,

New feature to add support to the "samesite" session cookie option via INI setting.

Note that when using the session_set_cookie_params I also updated the approach to use the alternative signature of that function which uses an associative array instead of individual vars (supported on PHP 7.3+)
https://www.php.net/manual/en/function.session-set-cookie-params.php

Using the array makes it easy to set the "samesite" option.

This shouldn't not affect older instances using < 7.3 unless we pull these changes to that particular repo again. And for new projects, I don't see us using PHP < 7.3 so I think we should be fine.
In any case, if you like to add backwards compatibility then we could probably add an additional check with "phpversion()"?

### === TESTING NOTES ===
You can pull the changes on an existing lovestack instance you may have or use staging Habitat https://staging.habitatmag.com/habitat_admin (http auth: habitat/habitat) - full instance details [here](https://howto.mugo.ca/Howto/Clients/Habitat-Magazine-and-The-Habitat-Group)
1. Go to backend login and confirm the samesite attribute of the eZ session cookie shows "None". 
2. Confirm the login works.
3. Update the site.ini using the specific override you need, like for instance `settings/override/site.ini.append.staging.php` and add `SameSite=Lax` or `SameSite=Strict`
4. Remove all cookies and reload, confirm the new samesite value is in place.

Considerations.
* 'None' is the default value
* When using None, 'Secure' must be true. If you set on the site.ini `CookieSecure=false` and you are on https, the login should not work because cookies can't be stored.

Let me know what you think about this small session cookie enhancement. 

Thanks!


carlosmauri:
Sorry wrong PR - closing ticket.

